PR #50, opened 2024-11-27
Non Linear Flow Examples

albertviilik:
An example of a flow, that is non-linear. 

```mermaid
graph TD
    start["Start"] -->|User chooses Good Cop| good_cop["Good Cop"]
    start -->|User chooses Bad Cop| bad_cop["Bad Cop"]

    good_cop -->|Finish conversation| end_conversation["End Conversation"]
    good_cop -->|Switch to Bad Cop| bad_cop

    bad_cop -->|Finish conversation| end_conversation
    bad_cop -->|Switch to Good Cop| good_cop

    end_conversation["End Conversation"]
```

Screenshot 2024-11-27 at 21 40 09

### DIFF
--- a/editor/examples/emotional_states.json
+++ b/editor/examples/emotional_states.json
@@ -1,0 +1,216 @@
+{
+  "initial_node": "start",
+  "nodes": {
+    "start": {
+      "messages": [
+        {
+          "role": "system",
+          "content": "You are a greeter who helps determine the user's initial emotional state. Ask them how they're feeling today and guide them to an appropriate emotional state."
+        }
+      ],
+      "functions": [
+        {
+          "type": "function",
+          "function": {
+            "name": "happy",
+            "description": "Transition to happy emotional state",
+            "parameters": {
+              "type": "object",
+              "properties": {}
+            }
+          }
+        },
+        {
+          "type": "function",
+          "function": {
+            "name": "sad",
+            "description": "Transition to sad emotional state",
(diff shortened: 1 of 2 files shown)


albertviilik:
Apologies for the _"make belief" (🍪 theft investigation)_ rather than _"real world use case"_ prompts/example. I tried to come up with something simple and game like to illustrate the principle. I love how this functions like a finite state machine, where you can define the nodes (states) and the transitions (functions) between them.

albertviilik:
emotion-fsm

markbackman:
This is great! Thanks for trying out the editor. @albertviilik have you tried running this with the Flows python package? If so, how does it work?

480jy:
Yes. I have, I love it! 

It's very nice. For example, if you run the `emotional_states.json` it really shows how the state transitions work. You say you're sad, it goes to sad. Ask to switch to happy, it goes there. Ask to end or go to neutral, it does that. And it shows in the logs that that transition has happened.

albertviilik:
Yes. I have, I love it!

It's very nice. For example, if you run the emotional_states.json it really shows how the state transitions work. You say you're sad, it goes to sad. Ask to switch to happy, it goes there. Ask to end or go to neutral, it does that. And it shows in the logs that that transition has happened.

markbackman:
Thanks for submitting this. I'm closing it since it uses static flows which are now deprecated.

PR #59, opened 2024-12-08
Update the editor to output a special __function:__ token for handlers, parse that token in the FlowManager

markbackman:
The Flows editor exports JSON, which will export function handlers as strings. To get this JSON to be runnable, we're adding a `__function__:` token to the handler, so the FlowManager can parse and register the function from the JSON. This is additive, as directly specifying the function name in the JSON is still supported.

This also comes with an update to the editor code to support adding the `__function__:` token. Updated the examples to reflect that change too.


markbackman:
@jamsea FYI

PR #65, opened 2024-12-17
Simplify the static flows API

markbackman:
### Added

- New `initial_system_message` field in `FlowConfig`, which allows setting a
  global system message for static flows.

### Changed

- Simplified FlowManager initialization by removing the need for manual context
  setup in static flows.
- Updated static examples to use the updated API.

### DIFF
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -61,56 +61,6 @@ class FlowManager:
         initialized (bool): Whether the manager has been initialized
         nodes (Dict[str, Dict]): Node configurations for static flows
         current_functions (Set[str]): Currently registered function names
-
-    Static Flow Example:
-        # Define a static flow configuration
-        flow_config = {
-            "initial_node": "greeting",
-            "initial_system_message": [{
-                "role": "system",
-                "content": "You are a helpful assistant"
-            }],
-            "nodes": {
-                "greeting": {
-                    "messages": [
-                        {"role": "system", "content": "Start by greeting the user"}
-                    ],
-                    "functions": [{
-                        "type": "function",
-                        "function": {
-                            "name": "collect_name",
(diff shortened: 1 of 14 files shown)


src/pipecat_flows/manager.py
```diff
@@ -167,25 +110,41 @@ def __init__(
...
         try:
-            # Set initial context with no tools
-            await self.task.queue_frame(LLMMessagesUpdateFrame(messages=initial_messages))
-            await self.task.queue_frame(LLMSetToolsFrame(tools=[]))
+            if self.nodes:  # Static flow
+                if initial_messages:
+                    logger.warning("Initial messages ignored for static flow configuration")
+
+                # Set initial system message if present
+                if hasattr(self, "initial_system_message"):
```
aconchillo, line 137: I think this could be:

```
if self.initial_system_message:
```

We should always have things (i.e. fields) defined, because this is kind of hiding what fields are defined or not depending on something else.

src/pipecat_flows/manager.py
```diff
@@ -155,6 +97,7 @@ def __init__(
         if flow_config:
             self.nodes = flow_config["nodes"]
             self.initial_node = flow_config["initial_node"]
+            self.initial_system_message = flow_config.get("initial_system_message", [])
             self.transition_callback = self._handle_static_transition
             logger.debug("Initialized in static mode")
         else:
```
aconchillo, line 103: in this else block we would have:

```
self.initial_system_message = None
```

aconchillo:
LGTM! Just made a small comment easy to fix.

aconchillo (approved):

src/pipecat_flows/manager.py
markbackman: Makes sense!

PR #68, opened 2024-12-19
Update FlowManager to support role and task messages

markbackman:
### Changed

- Nodes now have two message types to better delineate defining the role or
  persona of the bot from the task it needs to accomplish. The message types are:

  - `role_message`, which defines the personality or role of the bot
  - `task_message`, which defines the task to be completed for a given node

- `role_messages` can be defined for the initial node and then inherited by
  subsequent nodes. You can treat this as an LLM "system" message.

- Simplified FlowManager initialization by removing the need for manual context
  setup in both static and dynamic flows. Now, you need to create a `FlowManager`
  and initialize it to start the flow.
- All examples have been updated to align with the API changes.

---

## Flows editor

Also, the Flow Editor needed to be updated for the new message structure as well. This PR includes an update there, as well as updates to all editor JSON examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,16 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
-
-- New `initial_system_message` field in `FlowConfig`, which allows setting a
-  global system message for static flows.
-
 ### Changed
 
+- Nodes now have two message types to better delineate defining the role or
+  persona of the bot from the task it needs to accomplish. The message types are:
+
+  - `role_message`, which defines the personality or role of the bot
+  - `task_message`, which defines the task to be completed for a given node
+
+- `role_messages` can be defined for the initial node and then inherited by
+  subsequent nodes. You can treat this as an LLM "system" message.
+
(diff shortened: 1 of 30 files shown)


CHANGELOG.md
```diff
@@ -7,16 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Changed
 
-- New `initial_system_message` field in `FlowConfig`, which allows setting a
-  global system message for static flows.
+- Nodes now have two message types to better delineate defining the role or
+  persona of the bot from the task it needs to accomplish. The message types are:
 
-### Changed
+  - `role_message`, which defines the personality or role of the bot
+  - `task_message`, which defines the task to be completed for a given node
```
aconchillo, line 16: Should these say `role_messages` and `task_messages`? Or simply "Role message, which..." and "Task message, which...". Just mentioning since `role_message` is not an actual thing.

aconchillo (approved):

aconchillo:
Awesome! LGTM!

CHANGELOG.md
markbackman: Should be `...messages`. Thanks for catching that.

Issue #64, opened 2024-12-16
Anthropic module is required when not using it


Atharv24:
Hi, I added pipecat flows to my existing pipecat-ai workflow. I am using Silero, Deepgram, OpenAI and Daily for my project.

After installing pipecat-ai-flows, i tried running an example but got an import error for Anthropic.

markbackman:
Hey @Atharv24, do you mind sharing repro steps? I just tried the following and it worked fine.

1. Create a new venv:

```shell
python3 -m venv venv
source venv/bin/activate
```

2. Install pipecat:

```shell
pip install pipecat-ai
```

3. Create an .env file containing Deepgram, OpenAI, and Daily keys:

```shell
touch .env
```

4. Install pipecat extras for Daily, OpenAI, Deepgram, and Silero:

```shell
pip install "pipecat-ai[daily,openai,silero,deepgram]"
```

5. Install Pipecat Flows:

```shell
pip install pipecat-ai-flows
```

6. Created and installed dev-requirements.txt from Pipecat ([link](https://github.com/pipecat-ai/pipecat/blob/main/dev-requirements.txt)):

```shell
touch dev-requirements.txt
pip install -r dev-requirements.txt
```

7. Installed python-dotenv (This should be in the dev-requirements for Pipecat. I'll add that in a PR):

```shell
pip install python-dotenv
```

8. Ran this foundational example: https://github.com/pipecat-ai/pipecat/blob/main/examples/foundational/07c-interruptible-deepgram.py.

```shell
touch test.py
```

Pasted contents from 07c

```shell
python test.py
```

This ran as expected without needing Anthropic.

Can you share your steps?

gururise:
Same issue here.
```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/gene/Downloads/pipecat/examples/simple-chatbot/server/bot-gemini.py", line 55, in <module>
    from pipecat_flows import FlowArgs, FlowConfig, FlowManager, FlowResult
  File "/home/gene/Downloads/pipecat/.venv/lib/python3.12/site-packages/pipecat_flows/__init__.py", line 66, in <module>
    from .manager import FlowManager
  File "/home/gene/Downloads/pipecat/.venv/lib/python3.12/site-packages/pipecat_flows/manager.py", line 38, in <module>
    from pipecat.services.anthropic import AnthropicLLMService
[...]

PR #69, opened 2024-12-20
Lazy load the LLM modules to relax import requirements

markbackman:
Resolve this issue: https://github.com/pipecat-ai/pipecat-flows/issues/64.


aconchillo:
I think we should move everything the adapters do to Pipecat. Pipecat Flows shouldn't need to know about any specific LLM service, simply the `LLMService` class.

aconchillo (approved):

PR #71, opened 2024-12-22
Removed duplicate code from the NodeConfig class

dominikroblek:
Removed duplicate attribute declarations from the NodeConfig class.


markbackman:
Thanks for catching this.

markbackman (approved):

PR #72, opened 2024-12-29
Add state management access to static functions

markbackman:
Opening this draft with an experiment to make FlowManager state available.

TODOs:
- Revisit implementation to ensure this is developer friendly
- Write docs
- Revisit example for usability
- Get community feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,31 @@ There are two types of actions available:
 
 Learn more about built-in actions and defining your own action in the docs.
 
+#### State Management
+
+Flows can maintain state across nodes using two approaches:
+
+1. **Static Flows with State References**
+
+```python
+"task_messages": [
+    {
+        "role": "system",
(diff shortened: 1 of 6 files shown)


markbackman:
I posted this in Discord about this change:
> Static flows assume that the JSON is predetermined in advance; this means that your role_messages, task_messages, and function calls are all predefined as well. In a static flow, you can use handlers in your application logic for dynamic state.
>
> If you require dynamic changes of the role_messages, task_messages and function calls, then the dynamic flow is the way to go.
>
> I worked with someone at Daily through a static flow demo yesterday and kind of soured on the idea I proposed in the PR. I think I'm going to keep static flows truly static for now. If you need dynamic state, then dynamic flows are an option.

Closing this for now. If you need to dynamically change the flow, then I recommend using the dynamic flows.

PR #73, opened 2024-12-31
[FEATURE] Add arg to reset llm context while setting nodes

Atharv24:
Wanted to reset LLM context while setting new nodes. I was facing problems in long running conversations where large context created unwanted responses from the LLM.

### DIFF
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -324,7 +324,7 @@ class FlowManager:
                 if "transition_to" in decl:
                     del decl["transition_to"]
 
-    async def set_node(self, node_id: str, node_config: NodeConfig) -> None:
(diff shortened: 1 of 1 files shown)


markbackman:
@Atharv24 supporting a context reset is on my to do list. I plan to make resetting a configurable option that you can use if needed. I hope to get to it in the next week or so.

markbackman:
@Atharv24 I finally had a chance to look into this and I'm seeing that the function call results are removed from the context with this approach, which results in the LLM completion occurring with the necessary information. Were you able to work around this?

markbackman:
Starting work on this here: https://github.com/pipecat-ai/pipecat-flows/pull/88

markbackman:
@Atharv24 thanks for this submission. Check out #88.

Issue #67, opened 2024-12-19
LLM sends a message before state transition completes


Atharv24:
**Current Setup**
Interview Bot with Dynamic flow handling
initial node (greet user) -> call start_interview function (only a print statement here) -> call transition function (picks question to be asked from flow manager state) -> create and set node to question_node (specifies the question to be asked in LLM context)

**Expected Functionality**
- LLM greets the user
- Calls the start interview tool
- Tool runs successfully
- Transition to new question node with question to be asked in context
- LLM asks the question to user

**Actual Behaviour**
- LLM greets the user
- Calls the start interview tool
- Tool runs successfully
- Transition to new question node with question to be asked in context
- **LLM asks a question on its own**. Checking the LLM context at this point, I can see the new node context is not there. When I say one more line, next context that is printed shows the expected context messages.

As per my understanding, LLM sends a messages before the new node context is updated. Let me know if any code is needed from my end. I use the same format as in the dynamic flow example.

**UPDATE**: I added a 1 second sleep after setting the node and before the transition function ends, it works fine now.

Atharv24:
```
class InterviewStartResult(FlowResult):
    status: str

class AnswerCollectionResult(FlowResult):
    answer: str

class QuestionType(StrEnum):
    SUBJECTIVE = "SUBJECTIVE"
    CODING = "CODING"
    SUBJECTIVE_PROBING = "SUBJECTIVE_PROBING"

def create_initial_node(bot_prompt: str) -> NodeConfig:
    """Create the initial greeting node."""
    return {
        "messages": [
            {
                "role": "system",
                "content": bot_prompt,
            }
        ],
        "functions": [
            {
                "type": "function",
                "function": {
                    "name": "start_interview",
                    "handler": handle_start_interview,
                    "description": "Start the interview after candidate confirms",
                    "parameters": {"type": "object", "properties": {}},
                },
            }
        ],
    }

def create_question_node(
    question: dict, 
) -> NodeConfig:
    """Create node for asking questions with appropriate prompts."""

    question_content = (
        "Ask the candidate the following question:\n"
        f"QUESTION : {question['question']}\n\n"
        "Instructions:\n"
[...]

PR #75, opened 2025-01-14
Improve LLM completion timing for edge and node functions

markbackman:
Issue: https://github.com/pipecat-ai/pipecat-flows/issues/67.

This fix comes in two parts:
- Pipecat: https://github.com/pipecat-ai/pipecat/pull/970
- This Flows change

This PR resolves a timing/race condition issue that resulted in variable and unpredictable timing for how LLM completions occur. This change makes both node and edge function calls predictable in that:
- Edge function calls (which transition to a new node) result in an LLM completion after both the function call result and the next nodes messages are added to the context. This ensures that the LLM responds only once and with the necessary task and context.
- Node function calls (which happen within the existing node) result in an LLM completion once the function call result is added to the context (default Pipecat behavior).

⚠️ This change is breaking in that you need to pass the context_aggregator to the FlowManager at initialization. The benefit for this is warranted though as it results in LLM completions creating a natural and controllable conversation.

For reviewing, the code change is only in the manager.py. All of the other files are changes to align with the FlowManager updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@ All notable changes to **Pipecat Flows** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Updated `FlowManager` to more predictably handle function calls:
+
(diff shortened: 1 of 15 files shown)


markbackman:
Tests are failing because a newer version of Pipecat is required... I'll wait until the 0.0.53 release before merging this due to the dependency.

filipi87 (approved):
LGTM!

Issue #81, opened 2025-01-21
Pre-actions custom functions are required to be registered manually


tropxy:
Hey!

I was playing around with the pre-actions feature and I wanted to add a custom function that I want it to run, besides the `tts_say` and when I added this to the node configuration:

```python
"pre_actions": [
            {
                "type": "tts_say",
                "text": "Hi, welcome to AmpAssist, we are preparing to assist you. Please wait a moment.",
            },
            {
                "type": "test",
                "text": "This is a dummy pre function",
            },
        ],
```

my dummy test function wouldnt run

```python
async def test(message: FlowArgs) -> None:
    """Dummy Pre function
    message is a Dict thish in this case will be of the
    form
            {
                "type": "test",
                "text": "This is a dummy pre function",
            },
    """
    logger.debug(f"{message['text']}")

```

and the error reported was something like "function test was not registered".
Going over the inner code, I found out that there is a method publicly exposed to register custom functions/actions:
https://github.com/pipecat-ai/pipecat-flows/blob/main/src/pipecat_flows/manager.py#L140

The docstring made it pretty clear how to use it and the function ran with no issues after being manually registered.

My question would be if this is the intended user experience, and why not register the custom function described in the configuration node automatically, has done with the other functions outside of the pre-actions, done here:
https://github.com/pipecat-ai/pipecat-flows/blob/main/src/pipecat_flows/manager.py#L404 ?

Is it because these functions are different from the others as they do not trigger a transaction event and are not supposed to be functions that the LLM shall know about, as they are not on demand functions?

Code where the functions are registered in the LLM, for convenience and illustration purposes:
```python
# Let adapter format tools for provider
formatted_tools = self.adapter.format_functions(tools)

# Update LLM context
await self._update_llm_context(messages, formatted_tools)
```

If the flow to register a custom function under a pre-action is as done above, then at least I guess it should be good to provide some examples in the README or in the examples folder itself. 
I am happy to contribute btw ;)

markbackman:
This is an artifact of the original flows code that hasn't been updated. Flows is still a young project and as it gets more use, the dev ex and capabilities are getting refined. The very first iteration supported static flows only where LLM functions and actions needed to be manually registered.

Given the path that LLM functions have taken, I think auto-registration of actions makes sense to improve the developer experience.

> I am happy to contribute btw ;)
Sharing this feedback is really great! Very helpful to get feedback from someone building with Flows. This is what makes it better :)

[...]

PR #83, opened 2025-01-21
Mb/auto register custom actions

markbackman:
Response to: https://github.com/pipecat-ai/pipecat-flows/issues/81.

The aligns action registration with function call registration. Now both are done automatically.

This is a continued push for Flows to become more of an abstraction, removing the specifics of how underlying Pipecat and Pipecat-adjacent functionality works. Now, just define your pre or post actions in your node config to use them.


src/pipecat_flows/actions.py
```diff
@@ -32,6 +32,7 @@
 from pipecat.pipeline.task import PipelineTask
 
 from .exceptions import ActionError
+from .types import ActionConfig
```
filipi87, line 35: The code is much easier to understand with these types. Cool. 🙂

filipi87 (approved):
Looks good. 🚀

PR #87, opened 2025-01-28
Fix an issue where multiple, consecutive function calls result in repeat completions

markbackman:
This was a tricky one. In 0.0.11, I added constraint that completions would happen after either:
- Node function: just a function call
- Edge function: a function call and an update to the context

What I didn't account for is the case where the LLM will call a function multiple times consecutively. You can see this in the `food_ordering.py` example if you ask it for multiple orders. The LLM will execute the function call once per item asked. This results in calling the function multiple times.

For example, ordering "a large cheese and a medium pepperoni" triggers:
```python
2025-01-28 18:28:01.783 | INFO     | pipecat.processors.aggregators.openai_llm_context:call_function:200 - Calling function select_pizza_order with arguments {'size': 'medium', 'type': 'pepperoni'}
2025-01-28 18:28:01.784 | INFO     | pipecat.processors.aggregators.openai_llm_context:call_function:200 - Calling function select_pizza_order with arguments {'size': 'large', 'type': 'cheese'}
```

As a solution, I'm counting pending function calls. And, I'm waiting to transition to the next node until those function calls have yielded results. This is done through a counter. I went in this direction to not have to debounce or add latency.

### DIFF
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -300,9 +300,6 @@ class FlowManager:
                                 elif transition_callback:  # Dynamic flow
                                     logger.debug(f"Dynamic transition for: {name}")
                                     await transition_callback(args, self)
-                                # Reset counter after transition completes
-                                self._pending_function_calls = 0
-                                logger.debug("Reset pending function calls counter")
                             else:
                                 logger.debug(
                                     f"Skipping transition, {self._pending_function_calls} calls still pending"
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -259,6 +259,47 @@ class FlowManager:
 
         is_edge_function = bool(transition_to) or bool(transition_callback)
 
+        def decrease_pending_function_calls() -> None:
+            """Decrease the pending function calls counter if greater than zero."""
+            if self._pending_function_calls > 0:
+                self._pending_function_calls -= 1
+                logger.debug(
+                    f"Function call completed: {name} (remaining: {self._pending_function_calls})"
+                )
+
+        async def on_context_updated_edge(args: Dict[str, Any], result_callback: Callable) -> None:
+            """Handle context updates for edge functions with transitions."""
+            try:
+                decrease_pending_function_calls()
+
+                # Only process transition if this was the last pending call
+                if self._pending_function_calls == 0:
+                    if transition_to:  # Static flow
+                        logger.debug(f"Static transition to: {transition_to}")
+                        await self.set_node(transition_to, self.nodes[transition_to])
+                    elif transition_callback:  # Dynamic flow
+                        logger.debug(f"Dynamic transition for: {name}")
+                        await transition_callback(args, self)
+                    # Reset counter after transition completes
+                    self._pending_function_calls = 0
+                    logger.debug("Reset pending function calls counter")
+                else:
+                    logger.debug(
+                        f"Skipping transition, {self._pending_function_calls} calls still pending"
+                    )
+            except Exception as e:
+                logger.error(f"Error in transition: {str(e)}")
+                self._pending_function_calls = 0
+                await result_callback(
+                    {"status": "error", "error": str(e)},
+                    properties=None,  # Clear properties to prevent further callbacks
+                )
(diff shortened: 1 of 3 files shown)


src/pipecat_flows/manager.py
```diff
@@ -280,14 +287,29 @@ async def transition_func(
...
+                            if self._pending_function_calls == 0:
+                                if transition_to:  # Static flow
+                                    logger.debug(f"Static transition to: {transition_to}")
+                                    await self.set_node(transition_to, self.nodes[transition_to])
+                                elif transition_callback:  # Dynamic flow
+                                    logger.debug(f"Dynamic transition for: {name}")
+                                    await transition_callback(args, self)
+                                # Reset counter after transition completes
+                                self._pending_function_calls = 0
+                                logger.debug("Reset pending function calls counter")
```
filipi87, line 305: It is only goes inside this if, if we don't have any function calls pending. So, I guess we don't need this code ?

markbackman: Good point! Thank you.

markbackman: Ok, fixed.

src/pipecat_flows/manager.py
```diff
@@ -299,10 +321,15 @@ async def on_context_updated() -> None:
                     )
                     await result_callback(result, properties=properties)
```
filipi87, line 319: Do we need to decrease the counter here as well ?

filipi87: Maybe if we decreased here we wouldn't need to decrease inside the on_context_updated ? 🤔

markbackman: That might work.

I decrease on the on_context_updated callback because we know the function call result has been received and the function call result is in the context at that point in time. So, it's a safe place to know that the function call has completed; that's the signal for moving ahead with the transition. For that reason, I opted to also designate that as the place where we should decrement the counter and see if we can move ahead to the next node. WDYT?

filipi87: I see what you mean. What I am not so happy about is that one time we invoke it before calling `result_callback`, and another time, we invoke it inside the context.

What if we refactored, and did something like this:

```
                def decrease_pending_function_calls() -> None:
                    if self._pending_function_calls > 0:
                        self._pending_function_calls -= 1
                        logger.debug(
                            f"Node function completed: {name} (remaining: {self._pending_function_calls})"
                        )

                async def on_context_updated_edge() -> None:
                    try:
                        decrease_pending_function_calls()

                        # Only process transition if this was the last pending call
                        if self._pending_function_calls == 0:
                            if transition_to:  # Static flow
                                logger.debug(f"Static transition to: {transition_to}")
                                await self.set_node(transition_to, self.nodes[transition_to])
                            elif transition_callback:  # Dynamic flow
                                logger.debug(f"Dynamic transition for: {name}")
                                await transition_callback(args, self)
                        else:
                            logger.debug(
                                f"Skipping transition, {self._pending_function_calls} calls still pending"
                            )
                    except Exception as e:
                        logger.error(f"Error in transition: {str(e)}")
                        self._pending_function_calls = 0
                        await result_callback(
                            {"status": "error", "error": str(e)},
                            properties=None,  # Clear properties to prevent further callbacks
                        )
                        raise  # Re-raise to prevent further processing

                async def on_context_updated_node() -> None:
                    decrease_pending_function_calls()

                on_context_updated = on_context_updated_edge if is_edge_function else on_context_updated_node

                properties = FunctionCallResultProperties(
                    run_llm=False, on_context_updated=on_context_updated
                )
                await result_callback(result, properties=properties)
```

markbackman: That makes sense and is nice and clean. I've made an update along those lines. Can you take another look?

filipi87 (approved):
Looks good. 🙌
I’ve just noticed that the tests are failing. Do we need to update something, or is it not related to these changes?

markbackman:
> Looks good. 🙌 I’ve just noticed that the tests are failing. Do we need to update something, or is it not related to these changes?

I'll look at the test failure to see what's up. Thanks for the review!

PR #88, opened 2025-01-29
Add LLM context management strategies

markbackman:
Introduces context management strategies for Flows, enabling better control over conversation history and token use. This change adds three strategies for managing context during node transitions, including an AI-powered summarization option.

### Changes:

- Adds `ContextUpdateStrategy` enum with three options:
  - `APPEND` (default): Maintains full conversation history
  - `RESET`: Clears context while preserving critical function calls
  - `RESET_WITH_SUMMARY`: Resets context but includes AI-generated summary
- Updates `FlowManager` to support both global and per-node strategy configuration
- Implements provider-specific adapters for OpenAI, Anthropic, and Google
- Adds graceful fallback to `RESET` if summary generation fails in `RESET_WITH_SUMMARY` strategy

### Example usage:

```python
# Global strategy
flow_manager = FlowManager(
    task=task,
    llm=llm,
    context_aggregator=context_aggregator,
    context_strategy=ContextUpdateStrategy.APPEND
)

# Per-node strategy with summary
node_config = {
    "task_messages": [...],
    "functions": [...],
    "context_update_strategy": ContextUpdateConfig(
        strategy=ContextUpdateStrategy.RESET_WITH_SUMMARY,
        summary_prompt="Summarize the key points discussed so far."
    )
}
```

### Logic

When `APPEND` is specified:
- Messages are appended to the context throughout the conversation

When `RESET` is specified:
- Previous context is cleared
- Most recent function call/response pair is preserved
- New node messages are added
- Messages maintain proper ordering for each LLM provider

When `RESET_WITH_SUMMARY` is specified:
- Previous context is provided to LLM for summarization
- Generated summary is formatted appropriately for each provider
- New context starts fresh with summary + new node messages
- Falls back to RESET strategy if summary generation fails


markbackman:
Depends on https://github.com/pipecat-ai/pipecat/pull/1111 for Gemini to work.

CHANGELOG.md
```diff
@@ -5,6 +5,39 @@ All notable changes to **Pipecat Flows** will be documented in this file.
...
+    along with the new messages
+  - Strategies can be set globally or per-node
+  - Includes automatic fallback to RESET if summary generation fails
+
+Example usage:
+
+```python
+# Global strategy
+flow_manager = FlowManager(
+    context_strategy=ContextUpdateStrategy.APPEND
```
filipi87, line 27: It would be nice to have a consistent name for the field in both `FlowManager` and `node_config`. 
Maybe choose whether to use `context_strategy` or `context_update_strategy` in both cases ? WDYT ?

filipi87: Because I guess here we could also decide to use this, right ?
```
ContextUpdateConfig(
        strategy=ContextUpdateStrategy.RESET_WITH_SUMMARY,
        summary_prompt="Summarize the key points discussed so far."
    )
```

examples/static/patient_intake_anthropic.py
filipi87, line 1: I know that we have been organized this way for all the examples. But would'nt be better if instead of duplicating all the code, we have a base class, and them the specific examples for gemini and open ai just inheritance from this class providing what is specific for each one, which is basically the functions and the LLM service ?

And in the future, when the functions are in a standard format, the single difference will be the LLM service. 

I think that would make it a lot easier to improve and keep all the examples up to date. What do you think ?

CHANGELOG.md
```diff
@@ -5,6 +5,39 @@ All notable changes to **Pipecat Flows** will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added context update strategies to control how context is managed during node
+  transitions:
+  - `APPEND`: Add new messages to existing context (default behavior)
+  - `RESET`: Clear and replace context with new messages and most recent
```
filipi87, line 15: Why did we decide in this case to keep the recent function call results? I'm just curious to understand the motivation.

src/pipecat_flows/manager.py
```diff
@@ -77,6 +86,7 @@ def __init__(
         context_aggregator: Any,
         tts: Optional[Any] = None,
         flow_config: Optional[FlowConfig] = None,
+        context_strategy: Optional[ContextUpdateStrategy] = None,
```
filipi87, line 89: Shouldn't it be `ContextUpdateConfig` ? Because if I choose to use the `RESET_WITH_SUMMARY` strategy, how can I provide the `summary_prompt` in this case ?

src/pipecat_flows/manager.py
```diff
@@ -575,27 +589,79 @@ async def set_node(self, node_id: str, node_config: NodeConfig) -> None:
...
-            # Determine frame type based on whether this is the first node
+            update_config = strategy or ContextUpdateConfig(strategy=self._context_strategy)
+
+            if (
+                update_config.strategy == ContextUpdateStrategy.RESET_WITH_SUMMARY
+                and self._context_aggregator
+                and self._context_aggregator.user()._context.messages
+            ):
+                if not update_config.summary_prompt:
+                    raise FlowError("summary_prompt required for RESET_WITH_SUMMARY strategy")
```
filipi87, line 623: I guess it would break here if I had configured the flow with the strategy `RESET_WITH_SUMMARY`, and not provided a `ContextUpdateConfig` for the node ?

CHANGELOG.md
markbackman: Good idea. I'll go with `context_strategy` as it's more concise.

markbackman: I've made other updates to align names too.

examples/static/patient_intake_anthropic.py
markbackman: Yes! I'll do that as part of this. I was thinking about how this isn't a good pattern going forward. It would be nice to run all of the demos with whichever LLM provider.

CHANGELOG.md
markbackman: We keep all function calls up to the previous node message. The reason for this is that the function call may contain information required to complete the next step. For example, if you ask for the weather forecast for a city, the weather information is returned in the function call. If you eliminate the function call, the you eliminate the result from the function call.

And, on a node transition, the LLM completion only happens after the function call response and node messages are added. Keeping the function call info allows the LLM to include it in its response.

src/pipecat_flows/manager.py
markbackman: Yes, thank you for catching this.

src/pipecat_flows/manager.py
markbackman: I think it makes sense to get rid of that check all together. Instead, I have this _post_init__ validation:
```
@dataclass
class ContextStrategyConfig:
    """Configuration for context management.

    Attributes:
        strategy: Strategy to use for context management
        summary_prompt: Required prompt text when using RESET_WITH_SUMMARY
    """

    strategy: ContextStrategy
    summary_prompt: Optional[str] = None

    def __post_init__(self):
        """Validate configuration."""
        if self.strategy == ContextStrategy.RESET_WITH_SUMMARY and not self.summary_prompt:
            raise ValueError("summary_prompt is required when using RESET_WITH_SUMMARY strategy")
```

I think this is the better place to catch the error because falling back or providing a prompt just masks the problem. WDYT?

filipi87: Yep, this approach makes total sense. 🚀

CHANGELOG.md
filipi87: I see, it is a lot easier to understand now why we are keeping it. Thank you @markbackman

markbackman:
@filipi87 this one should be ready for review now. I've fixed the items in your comments, added tests, and docs (separate [PR](https://github.com/pipecat-ai/docs/pull/113)).

I'm going to improve the examples, per your [comment](https://github.com/pipecat-ai/pipecat-flows/pull/88#discussion_r1937166925), in a separate PR.

filipi87 (approved):
Looks good. Pretty nice feature. 🙌

PR #90, opened 2025-02-01
tts_say action should use TTSSpeakFrame

markbackman:
This was an old to do, which is now unblocked. It's an internal change to push a `TTSSpeakFrame` for a `tts_say` action instead of using the `tts.say()` action.


filipi87 (approved):
🙌🚀

PR #91, opened 2025-02-01
Add a `get_current_context` function to access the latest context

markbackman:
This request came up in a conversation with a developer. It could be handy to access the context directly since the FlowManager has access to it.


filipi87 (approved):
Looks good.

PR #92, opened 2025-02-04
Add a dynamic restaurant reservation demo

markbackman:
This demo shows how to use Flows to make routing decisions based on third-party API results. Key points:

1. Function calls are structured through LLM function parameters:
```python
"parameters": {
    "type": "object",
    "properties": {
        "time": {
            "type": "string",
            "pattern": "^([5-9]|10):00 PM$",
            "description": "Reservation time (e.g., '6:00 PM')",
        },
        "party_size": {"type": "integer"},
    },
    "required": ["time", "party_size"],
}
```

2. Return values are structured using TypedDict as a `FlowResult`:
```python
class TimeResult(FlowResult):
    status: str
    time: str
    available: bool
    alternative_times: list[str]
```

The `FlowResult` gets serialized to JSON and added to the LLM's context as a "tool" message, making the result available for:
- The LLM's next response
- Transition decisions in our flow


CHANGELOG.md
```diff
@@ -50,6 +50,14 @@ Example usage:
 context = flow_manager.get_current_context()
 ```
 
+### Changed
+
+- Function results are now automatically stored in flow_manager.state using
```
kompfner, line 55: is there no way to pass the `FlowResult` from the function call in as a argument to the `transition_callback`? like:

```python
def handle_quote_calculation(args: Dict, flow_manager: FlowManager, function_result: FlowResult = None)
```

that feels a bit more ergonomic, and sidesteps the problem of using a name in `flow_manager.state`, which—if i understand correctly—is meant to be a "user space" thing.

of course, the above approach assumes that there's only one function call preceding a `transition_callback` call. which is true, right?

markbackman: The other option I considered was passing the result to the transition callback as an arg. To not introduce a breaking change, we'd have to support both transition_callback function signatures. That's something we can definitely do if you think it's more ergonomic.

examples/dynamic/restaurant_reservation.py
```diff
@@ -0,0 +1,346 @@
...
+    flow_manager.state["party_size"] = args["size"]
+    await flow_manager.set_node("get_time", create_time_selection_node())
+
+
+async def handle_availability_check(args: Dict, flow_manager: FlowManager):
+    """Handle availability check result and transition based on availability."""
+    # Store reservation details in flow state
+    flow_manager.state["requested_time"] = args["time"]
+
+    # Get result from flow state instead of calling function again
```
kompfner, line 116: nice! bonus: would also love to see `handle_quote_calculation()` updated in the `insurance_*` examples, as that's where we first noticed an obvious opportunity for avoiding a redundant function call.

(side note: not only does this new pattern avoid _redundant_ function calls, but in the case of nondeterministic or mutating function call results, this new pattern is the _only_ way to achieve what you want)

markbackman: Yup! Good call. I can take that on as an action. Makes sense to re-write that given this new pattern.

markbackman:
@kompfner I've updated this to match what we discussed. Check it out.

filipi87 (approved):
Nice improvement. 🚀

PR #93, opened 2025-02-04
Make transport accessible from node custom action handlers. This invo…

kompfner:
…lves two changes:

- Hang the transport off of the `FlowManager`
- Pass the `FlowManager` to the action handlers

### DIFF
--- a/src/pipecat_flows/actions.py
+++ b/src/pipecat_flows/actions.py
@@ -109,16 +109,20 @@ class ActionManager:
                 raise ActionError(f"No handler registered for action type: {action_type}")
 
             try:
-                # Check if the handler can handle a second flow_manager arg (i.e. if it has multiple or variable positional args)
-                handler_positional_arg_count = handler.__code__.co_argcount
-                if not isinstance(handler_positional_arg_count, int):
-                    # (Here we deal with a bug in python 3.10's Mock class where arg count will be a Mock; assume handler can handle flow_manager arg)
-                    can_handle_flow_manager_arg = True
-                else:
+                # Determine if handler can accept flow_manager argument by inspecting its signature
+                # Handlers can either take (action) or (action, flow_manager)
+                try:
+                    handler_positional_arg_count = handler.__code__.co_argcount
                     if inspect.ismethod(handler) and handler_positional_arg_count > 0:
                         # adjust for `self` being the first arg
                         handler_positional_arg_count -= 1
-                    can_handle_flow_manager_arg = handler_positional_arg_count > 1 or handler.__code__.co_flags & 0x04 # *args
+                    can_handle_flow_manager_arg = handler_positional_arg_count > 1 or handler.__code__.co_flags & 0x04
+                except AttributeError:
+                    logger.warning(
+                        f"Unable to determine handler signature for action type '{action_type}', "
+                        "falling back to legacy single-parameter call"
+                    )
+                    can_handle_flow_manager_arg = False
 
                 # Invoke handler appropriately, with async and flow_manager arg as needed
                 if can_handle_flow_manager_arg:
--- a/src/pipecat_flows/actions.py
+++ b/src/pipecat_flows/actions.py
@@ -109,12 +109,14 @@ class ActionManager:
                 raise ActionError(f"No handler registered for action type: {action_type}")
 
             try:
-                handler_signature = inspect.signature(handler)
-                if len(handler_signature.parameters) > 1 or any(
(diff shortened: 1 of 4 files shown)


markbackman:
Curious why you're passing the entire flow manager and not just the transport. I suppose that adds more flexibility in the future.

kompfner:
Stay tuned for unit test changes.

src/pipecat_flows/actions.py
```diff
@@ -51,16 +51,16 @@ class ActionManager:
     Custom actions can be registered using register_action().
     """
 
-    def __init__(self, task: PipelineTask, tts=None):
```
kompfner, line 54: do we think we should _also_ get rid of the `task` parameter here, since `task` can be publicly accessed off of `flow_manager`?

kompfner: hm, i'm actually now leaning more towards the second option i presented above. it's good to have a record, in the constructor, of what `ActionManager` actually needs to directly access...

src/pipecat_flows/actions.py
```diff
@@ -106,10 +108,17 @@ async def execute_actions(self, actions: Optional[List[ActionConfig]]) -> None:
                 raise ActionError(f"No handler registered for action type: {action_type}")
 
             try:
-                if asyncio.iscoroutinefunction(handler):
-                    await handler(action)
+                if 'flow_manager' in handler.__code__.co_varnames:
```
kompfner, line 111: this is wrong. fixing

kompfner:
Note to self: this will require docs updates

src/pipecat_flows/actions.py
```diff
@@ -106,10 +109,21 @@ async def execute_actions(self, actions: Optional[List[ActionConfig]]) -> None:
                 raise ActionError(f"No handler registered for action type: {action_type}")
 
             try:
-                if asyncio.iscoroutinefunction(handler):
-                    await handler(action)
+                handler_signature = inspect.signature(handler)
```
kompfner, line 112: grr, this looks like it doesn't work with `AsyncMock` in python 3.10 (which our CI uses), but it does work in python 3.13. thoughts on what to do here? what's the min version of python that we publicly claim to support?

kompfner: ah just found that we do `requires-python = ">=3.10"`

kompfner: It did not work. Maybe in those other cases, we never ended up trying to get the signature of a `Mock`

kompfner: It's OK though—I figured it out. What's in the PR now does work.

src/pipecat_flows/actions.py
```diff
@@ -106,10 +109,28 @@ async def execute_actions(self, actions: Optional[List[ActionConfig]]) -> None:
                 raise ActionError(f"No handler registered for action type: {action_type}")
 
             try:
-                if asyncio.iscoroutinefunction(handler):
-                    await handler(action)
+                # Check if the handler can handle a second flow_manager arg (i.e. if it has multiple or variable positional args)
```
markbackman, line 112: Sorry, I didn't think about this when we talked. Isn't the `ActionManager` class, effectively on the whole, a private class? Also, I think `execute_actions` is also a private function. Maybe we should:
- Rename to `_execute_actions`
- Ignore the breaking change since this isn't exposed. This would help to simplify the code.

WDYT?

src/pipecat_flows/actions.py
markbackman: We decided to keep the FlowManager class!

src/pipecat_flows/actions.py
kompfner: If a customer was previously using a _custom_ pre- or post-action that took only one argument, here we're ensuring backward compatibility with that.

kompfner: Also: happy to rename to `_execute_actions` if you'd like, but was not treating any refactoring/renaming parto f the scope of these changes.

markbackman: Sorry, I've been distracted with core Pipecat work. Of course, this makes sense!

markbackman (approved):
@kompfner I finally carved out time to look at this. I made a few small changes:

- I made the types clearer by using type aliases for the two different signatures. This should improve type hinting for IDEs
- I saw the test issue you encountered. I wanted to simplify that. I ended up using action functions in the test_custom_action test instead of mocks. This allows us to simplify the code. I think this is for the best because it should be easier to maintain.

Anyway, tests are passing and I think this is good to go!

kompfner:
Nice improvements, @markbackman! Merging

PR #95, opened 2025-02-08
Revert "tts_say action should use TTSSpeakFrame"

markbackman:
- Reverting https://github.com/pipecat-ai/pipecat-flows/pull/90/commits/45ab5bfda7097c3999db21abb756f0fefde81e88
- Bumping to 0.0.14 to release this hotfix

I was mistaken. Before this change is made, we need to fix an ordering issue in the core Pipecat code to correctly handle the timing of `TTSSpeakFrame`s vs function calling. It seems the `TTSSpeakFrame` timing is too variable to rely on. We still need the unqueued `tts.say()` action.


aconchillo (approved):

aconchillo:
LGTM

PR #97, opened 2025-02-20
Fix the logic for creating LLM adapters.

kompfner:
As it was implemented, it was requiring devs to specify as pip requirements *all* optional LLM dependencies—google, anthropic, openai—regardless of which subset the dev actually needed.

The reason was that attempts to `import` from `pipecat.services.<llm>` would actually result in a different exception raised (`Exception(f"Missing module: {e}")`) than the one that was being handled (`ImportError`).

But something else was amiss here. By even *attempting* to `import` from an LLM service module we didn't need—say, `pipecat.services.anthropic`—we would inadvertently trigger some scary logger errors saying that the module and corresponding API key were missing, even though they're not needed.

It's worth noting there's no safety-related reason to import any LLM service modules here: if the dev has forgotten to install the appropriate dependency, they'll hit the error when trying to create the LLM service—e.g. `AnthropicLLMService`—in the first place.

### DIFF
--- a/src/pipecat_flows/adapters.py
+++ b/src/pipecat_flows/adapters.py
@@ -540,8 +540,8 @@ class GeminiAdapter(LLMAdapter):
 def create_adapter(llm) -> LLMAdapter:
     """Create appropriate adapter based on LLM service type.
 
-    Uses lazy imports to avoid requiring all provider dependencies at runtime.
-    Only the dependency for the chosen provider needs to be installed.
+    Use "stringy" checks to check the LLM service type instead of trying to import each provider 
+    dependency, as those will produce scary errors on the console even when they're not needed.
 
     Args:
         llm: LLM service instance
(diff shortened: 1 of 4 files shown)


src/pipecat_flows/adapters.py
```diff
@@ -540,8 +540,8 @@ async def generate_summary(
 def create_adapter(llm) -> LLMAdapter:
     """Create appropriate adapter based on LLM service type.
 
-    Uses lazy imports to avoid requiring all provider dependencies at runtime.
-    Only the dependency for the chosen provider needs to be installed.
+    Use "stringy" checks to check the LLM service type instead of trying to import each provider
```
markbackman, line 543: Maybe change this copy to:

```suggestion
    Uses string-based type checking to avoid importing unnecessary dependencies.
```

src/pipecat_flows/adapters.py
```diff
@@ -552,35 +552,14 @@ def create_adapter(llm) -> LLMAdapter:
...
-    # Try Google
-    try:
-        from pipecat.services.google import GoogleLLMService
-
-        if isinstance(llm, GoogleLLMService):
-            logger.debug("Creating Google adapter")
-            return GeminiAdapter()
-    except ImportError as e:
-        logger.debug(f"Google import failed: {e}")
+    if type(llm).__name__ == "OpenAILLMService":
```
markbackman, line 555: Nit: perhaps add logging and simplify slightly with:

```
llm_type = type(llm).__name__
    
if llm_type == "OpenAILLMService":
    logger.debug("Creating OpenAI adapter")
    return OpenAIAdapter()
        
if llm_type == "AnthropicLLMService":
    logger.debug("Creating Anthropic adapter")
    return AnthropicAdapter()
        
if llm_type == "GoogleLLMService":
    logger.debug("Creating Google adapter")
    return GeminiAdapter()
```

markbackman (approved):
LGTM! Just two nits in the comments.

src/pipecat_flows/adapters.py
kompfner: Actually, on second thought, your suggested wording is enough. I like it.

PR #101, opened 2025-02-26
Add support for function call start_callback

markbackman:
Pipecat's LLMServices support a `start_callback` before a function call runs. This callback allows you to execute functionality before the LLM function call. It's handy when you want to push a frame or run functionality before the function call takes place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added support for function `start_callback` to provide immediate feedback
+  before function execution.
+
+Example usage:
+
+```python
+# Define a start callback
+async def before_check_availability(function_name: str, llm: Any, context: Any) -> None:
+    """Provide user feedback during API calls."""
+    await llm.push_frame(TTSSpeakFrame("Let me check that for you..."))
+
+# Register in function configuration
+{
+    "type": "function",
+    "function": {
+        "name": "check_availability",
+        "handler": check_availability_handler,
+        "start_callback": before_check_availability,  # Add start callback
+        "description": "Check reservation availability",
+        "parameters": {...}
+    }
+}
+```
+
+### Other
+
+- Updated restaurant_reservation.py example to demonstrate the start callback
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added support for function `start_callback` to provide immediate feedback
```
kompfner, line 12: `start_callback` feels like an odd name to me, but i see now that it was already established as a name with `pipecat` 👍

markbackman: Yeah, it's a weird name, but it's the same name used in Pipecat...

CHANGELOG.md
```diff
@@ -7,6 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  before function execution.
+
+Example usage:
+
+```python
+# Define a start callback
+async def before_check_availability(function_name: str, llm: Any, context: Any) -> None:
```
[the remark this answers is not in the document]
kompfner: If we could choose the args we wanted for this callback rather than just go with what `pipecat`'s `LLMService` exposes, what would we want? I feel like we might want to provide the `FlowManager` wholesale, like we decided to do for pre-/post-actions.

src/pipecat_flows/manager.py
```diff
@@ -444,12 +445,29 @@ async def _register_function(
                     func_name = handler.split(":")[1]
                     handler = self._lookup_function(func_name)
 
+                # Create the wrapper for the start_callback that passes flow_manager
+                async def start_callback_wrapper(
+                    function_name: str, llm: Any, context: Any
+                ) -> None:
+                    if start_callback:
+                        try:
+                            # Pass flow_manager instead of llm and context
```
filipi87, line 454: Ok, I was trying to figure out why we had created a wrapper for it. I can see it now. 👍

markbackman: Yup, this was in response to Paul's comment. We wanted to avoid using the default args from Pipecat and instead offer the FlowManager as an arg.

PR #102, opened 2025-02-28
Warm transfer demo

kompfner:
Plan:
* First get this demo working _without_ dial-in/dial-out. Just rely on the "customer" and the "agent" joining the Daily room directly in their browser.
* Then implement dial-in/dial-out—maybe in another demo? The complication is that getting dial-in/dial-out set up is pretty nontrivial for the user.


examples/static/warm_transfer.py
```diff
@@ -0,0 +1,526 @@
...
+            remote_participants={
+                customer_participant_id: {
+                    "permissions": {
+                        "canSend": [],
+                    }
+                }
+            }
+        )
+
+    # TODO: there's no way in Daily to update remote participant's subscriptions :(
```
kompfner, line 148: This is the part that's waiting on incoming Daily-side work

kompfner:
Note to self: do not rebase `warm-transfer-demo` on top of `function-action-type` until [the issue](https://github.com/pipecat-ai/pipecat-flows/pull/107#discussion_r1996138097) with the last commit of `function-action-type` is fixed. Rebasing will cause the demo to stop working.

markbackman:
This is really great! While the logic is complex, it's relatively easy to follow along. (I wonder if we should think about adding conveniences to make controlling remote participants easier, especially for a customer/agent/bot scenario. That's for another PR though.)

I've noticed that sometimes the `canReceive()` change works and other times it doesn't. Not sure if this is relevant, but I see the following errors when the command runs:
```
{"timestamp":"2025-03-27T15:00:45.513169Z","level":"ERROR","fields":{"message":"no subscription for consumer: ConsumerId(\"0daee4f5-5d5d-41cb-81e6-9bbb55c5d8e7\")"},"target":"daily_core::call_manager::events::from_sfu::soup_consumer_closed"}
{"timestamp":"2025-03-27T15:00:45.562691Z","level":"ERROR","fields":{"message":"Failed to close consumer: ConsumerNoLongerExists(ConsumerId(\"0daee4f5-5d5d-41cb-81e6-9bbb55c5d8e7\"))"},"target":"daily_core::call_manager::events::subscription::common"}
```

I'm running Pipecat 0.0.61 and Flows 0.0.16.

One minor UX improvement might be to print the agent URL when it's time for the agent to join, rather than at the start of the session.

Other that, this is really on the right track. Very impressive! 👏

examples/dynamic/warm_transfer.py
```diff
@@ -0,0 +1,667 @@
...
+                audio_out_enabled=True,
+                vad_enabled=True,
+                vad_analyzer=SileroVADAnalyzer(),
+                vad_audio_passthrough=True,
+            ),
+        )
+        stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))
+        tts = CartesiaTTSService(
+            api_key=os.getenv("CARTESIA_API_KEY"),
+            voice_id="820a3788-2b37-4d21-847a-b65d8a68c99a",  # Salesman
```
markbackman, line 574: I love this voice, but it has a lot of problems in the new Cartesia model. You might want to try the Newsman: `d46abd1d-2d02-43e8-819f-51fb652c1c61`.

examples/dynamic/warm_transfer.py
```diff
@@ -0,0 +1,667 @@
...
+            """
+            user_id = participant.get("info", {}).get("userId")
+            if user_id == "agent" and flow_manager.current_node == "transferring_to_human_agent":
+                await start_human_agent_interaction(flow_manager=flow_manager)
+
+        @transport.event_handler("on_participant_left")
+        async def on_participant_left(
+            transport: DailyTransport, participant: Dict[str, Any], reason: str
+        ):
+            # TODO: update this logic so that if the customer participant has left early while on hold, inform the human agent; will require a new node?
```
markbackman, line 633: Just flagging this TODO. I think it's fine to keep it simple. It's just a demo, so no all cases need to be handled.

examples/dynamic/warm_transfer.py
kompfner: Yes! I've noticed so many weird audio quirks, too, like dogs barking in the background, weird laughter, and sudden changes in tone.

markbackman: I notified Cartesia about this one today. Hopefully they fix it. It's my favorite voice 😢

examples/dynamic/warm_transfer.py
```diff
@@ -0,0 +1,667 @@
...
+                - Use the check_store_location_and_hours_of_operation function to check store location and hours of operation to provide to the customer
+                - Use the start_order function to begin placing an order on the customer's behalf
+
+                If the customer wants to end the conversation, call the end_customer_conversation function.
+                """,
+            }
+        ],
+        functions=[
+            {
+                "type": "function",
```
markbackman, line 287: Random thought: should we use the new FlowsFunctionSchema for this? That would make it easier if someone wanted to try this example with a different LLM.

kompfner:
> I see the following errors when the command runs

These errors are a red herring—they occur no matter what. We're revoking the customer's `canSend` permission while they're on hold to prevent the bot from hearing anything they say. daily-python apparently isn't very gracefully handling the customer's tracks going away (i.e. the mediasoup consumer suddenly closing on the server side).

kompfner:
> I've noticed that sometimes the `canReceive()` change works and other times it doesn't

I'll keep testing, with an eye towards this. I've noticed occasional misbehaviors (especially when using Gemini v OpenAI)...I really don't think it has to do with the `canReceive` update mechanism not working when it's triggered but rather with the `start_order` function call actually getting triggered properly (I've occasionally seen the LLM not invoke it as it should, which suggests a prompt issue) and/or the post-action `post_transferring_to_human_agent` getting triggered at the appropriate time (but not sure about this one).

examples/dynamic/warm_transfer.py
kompfner: Done

examples/dynamic/warm_transfer.py
kompfner: For now, switched to Newsman. Thanks for that suggestion.

kompfner:
> I've noticed that sometimes the `canReceive()` change works and other times it doesn't

Hmm, OK, what I _think_ we're seeing reasonably often is that the bot's speech (triggered by `task_messages`) seems to be cut off by a `post_action`.

When we're the customer, a `"function"`-type `post_action` causes the bot to become inaudible too early (due to `canReceive` permission being revoked).

When we're the agent, the `"end_conversation"` `post_action` causes the bot to leave too early.

kompfner:
> Hmm, OK, what I _think_ we're seeing reasonably often is that the bot's speech (triggered by `task_messages`) seems to be cut off by a `post_action`.

I think the refactor in [this commit](https://github.com/pipecat-ai/pipecat-flows/pull/102/commits/7da759ba5885de54d65069aea6029219d9a3efed) made the issue a lot more prominent for some reason...

kompfner:
> Going to try to create a minimal repro for this.

Confirmed! Interrupting the bot in node A causes `post_actions` to start prematurely in node B, executing before the bot finishes speaking the new node's content.

examples/dynamic/warm_transfer.py
kompfner: Done.

examples/dynamic/hold_music/README.md
```diff
@@ -0,0 +1,7 @@
+# Hold Music Player
```
markbackman, line 1: Do we want the `hold_music` folder to be within an `assets` directory within `examples`? Maybe `examples/assets`?

kompfner: Good idea!

PR #106, opened 2025-03-13
Pass the FlowManager to function handlers

markbackman:
@kompfner this extends the pattern we discussed where the `FlowManager` should be generally available to different aspects of Flows. Let me know what you think.

I'm looking to preserve backwards compatibility for just providing `FlowArgs`, too.


kompfner (approved):
Nice improvement!

PR #107, opened 2025-03-13
Add a new "function" action type, which queues a function to run "inl…

kompfner:
…ine" in the pipeline.

This is a new way of creating "custom" actions, and is what we should recommend unless someone intends to use the customer action to queue a frame themselves.

✅ This PR first requires [this change](https://github.com/pipecat-ai/pipecat/pull/1368) to land and be published, and for pipecat-flow's dependency on pipecat to be bumped to that published version.

✅ This PR is waiting on [this fix](https://github.com/pipecat-ai/pipecat/pull/1442) to land be published, and for pipecat-flow's dependency on pipecat to be bumped to that published version. The change has landed, but has not yet been published.

### DIFF
--- a/src/pipecat_flows/actions.py
+++ b/src/pipecat_flows/actions.py
@@ -22,6 +22,7 @@ Actions are used to perform side effects during conversations, such as:
 """
 
 import asyncio
+from dataclasses import dataclass
 import inspect
 from typing import Callable, Dict, List, Optional
 
@@ -31,9 +32,16 @@ from pipecat.frames.frames import (
     TTSSpeakFrame,
 )
 from pipecat.pipeline.task import PipelineTask
+from pipecat.frames.frames import ControlFrame
 
 from .exceptions import ActionError
-from .types import ActionConfig
+from .types import ActionConfig, FlowActionHandler
+
+
+@dataclass
+class FunctionActionFrame(ControlFrame):
+    action: dict
+    function: FlowActionHandler
 
 
 class ActionManager:
@@ -68,6 +76,12 @@ class ActionManager:
         # Register built-in actions
         self._register_action("tts_say", self._handle_tts_action)
         self._register_action("end_conversation", self._handle_end_action)
+        self._register_action("function", self._handle_function_action)
+
+        @task.event_handler("on_frame_reached_downstream")
+        async def on_frame_reached_downstream(task, frame):
+            if isinstance(frame, FunctionActionFrame):
+                await frame.function(frame.action, flow_manager)
 
     def _register_action(self, action_type: str, handler: Callable) -> None:
         """Register a handler for a specific action type.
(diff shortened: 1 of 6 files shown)


src/pipecat_flows/actions.py
```diff
@@ -68,6 +76,12 @@ def __init__(self, task: PipelineTask, flow_manager: "FlowManager", tts=None):
         # Register built-in actions
         self._register_action("tts_say", self._handle_tts_action)
         self._register_action("end_conversation", self._handle_end_action)
+        self._register_action("function", self._handle_function_action)
+
+        @self.task.event_handler("on_frame_reached_downstream")
```
aconchillo, line 81: you can also refer to `@task.event_handler` directory since it's available here.

kompfner:
@aconchillo I'm a bit concerned about how [the change to task-ify all events](https://github.com/pipecat-ai/pipecat/pull/1376/files#r1995655047) could impact this feature...let me know what you think!

src/pipecat_flows/actions.py
```diff
@@ -176,3 +194,23 @@ async def _handle_end_action(self, action: dict) -> None:
...
+                Required 'handler' key containing the function to execute.
+        """
+        handler = action.get("handler")
+        if not handler:
+            logger.error("Function action missing 'handler' field")
+            return
+        # the reason we're queueing a frame here is to ensure it happens after bot turn is over in 
+        # post_actions
+        await self.task.queue_frame(FunctionActionFrame(action=action, function=handler))
+        await self.function_finished_event.wait()
```
kompfner, line 215: A note to reviewers: as-is, this line of code—introduced in [this commit](https://github.com/pipecat-ai/pipecat-flows/pull/107/commits/ee88d866a4279a9e02c664f08bd7fc3e5da2a2e4)—will hang forever.

Why?

Because pipecat-flows node transitions actually occur _entirely within the LLM service's callback function_, which runs in the LLM service's `__input_frame_task_handler()`. So `await`ing here actually stops the LLM service from processing any new frames, meaning that the queued `FunctionActionFrame` never makes it through the pipeline, and the function is never run. The pipeline grinds to a halt. It's deadlocked.

kompfner: After speaking with @aconchillo, it sounds like he has plans to make function calls in pipecat occur in separate tasks, so that long-running functions don't stall frame processing like they do today. This would alleviate the problem described above. I'd think we would need to be pretty careful with this change, though...seems like the kind of thing that could very easily have unintended side-effects...

kompfner: The function-calls-as-tasks change landed, but there's [another one](https://github.com/pipecat-ai/pipecat/pull/1442) that this PR is now waiting on.

kompfner: Looks like that PR still doesn't solve the pipeline deadlock problem...

kompfner: With the latest fixes in that PR, it _does_ solve the deadlock problem.

markbackman (approved):
LGTM.

Looking forward to seeing this in a demo. Sounds like the warm transfer will be the demo. I'll look at that in the AM.

aconchillo (approved):

PR #113, opened 2025-03-19
Revert: Add support for function call start_callback

markbackman:
`start_callback` is deprecated in Pipecat, so I'm removing it before it gets released.


filipi87 (approved):
LGTM

PR #115, opened 2025-03-25
Add support for FlowsFunctionSchema

markbackman:
Previously, Flows had it's own adapter classes. Now that Pipecat has classes for function calls, this PR brings that support to Flows. You can now:
- Specify a function call using the native LLM format
- Specify a function call using the FunctionSchema format

Note: Since Flows overloads the function call format to add a function `handler` and transition config in `transition_to` and `transition_callback`, we have to specify a new type: `FlowsFunctionSchema`. This is inspired by the `FunctionSchema`, but doesn't subclass it due to inheritance complexities across the packages. With this developers can provide the `FlowsFunctionSchema` format in their `NodeConfig`.

This PR aims to simplify the adapter logic, while still maintaining backwards compatibility for native LLM function call formats.

I'm bumping the minimum `pipecat-ai` version to 0.0.60 to support this new feature.


src/pipecat_flows/manager.py
```diff
@@ -482,50 +460,6 @@ async def _register_function(
                 logger.error(f"Failed to register function {name}: {str(e)}")
                 raise FlowError(f"Function registration failed: {str(e)}") from e
 
-    def _remove_handlers(self, tool_config: Dict[str, Any]) -> None:
```
markbackman, line 485: No longer need to remove, as this is done in the FunctionSchema conversion.

src/pipecat_flows/manager.py
```diff
@@ -482,50 +460,6 @@ async def _register_function(
...
-        if "function" in tool_config and "handler" in tool_config["function"]:
-            del tool_config["function"]["handler"]
-        elif "handler" in tool_config:
-            del tool_config["handler"]
-        elif "function_declarations" in tool_config:
-            for decl in tool_config["function_declarations"]:
-                if "handler" in decl:
-                    del decl["handler"]
-
-    def _remove_transition_info(self, tool_config: Dict[str, Any]) -> None:
```
markbackman, line 500: Same here: No longer need to remove, as this is done in the FunctionSchema conversion.

src/pipecat_flows/manager.py
```diff
@@ -259,29 +260,6 @@ async def _call_handler(self, handler: FunctionHandler, args: FlowArgs) -> FlowR
             # Modern handler with args and flow_manager
             return await handler(args, self)
 
-    async def _handle_static_transition(
```
markbackman, line 262: This was dead code I'm removing.

tests/test_manager.py
```diff
@@ -689,42 +697,6 @@ async def test_update_llm_context_error_handling(self):
                 messages=[{"role": "system", "content": "Test"}], functions=[]
             )
 
-    async def test_handler_removal_all_formats(self):
```
markbackman, line 692: Related code removed.

tests/test_manager.py
```diff
@@ -1177,3 +1160,81 @@ async def modern_handler(args: FlowArgs, flow_mgr: FlowManager) -> FlowResult:
         self.assertTrue(correct_flow_manager)
         self.assertEqual(result["args_received"]["test"], "value")
         self.assertTrue(result["has_flow_manager"])
+
+    async def test_pending_function_calls_counter(self):
```
markbackman, line 1164: This test was missing from a previous PR. I don't love this test; I had to rely on Claude to help me figure out a way to test a function call executing. I think it's OK for now.

markbackman: Removing this because it's problematic with CI. We'll have to revisit...

examples/dynamic/insurance_gemini.py
```diff
@@ -176,7 +176,59 @@ async def handle_end_quote(_: Dict, result: FlowResult, flow_manager: FlowManage
     await flow_manager.set_node("end", create_end_node())
 
 
-# Node configurations
+# Function definitions using FlowsFunctionSchema
```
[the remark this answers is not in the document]
kompfner: Or even keep them instantiated right inline in the node definition, where they were before, rather than pulling them elsewhere and giving them `*_schema` names.

src/pipecat_flows/adapters.py
```diff
@@ -37,41 +43,85 @@ class LLMAdapter(ABC):
...
-    @abstractmethod
-    def get_function_name(self, function_def: Dict[str, Any]) -> str:
-        """Extract function name from provider-specific function definition."""
-        pass
+    def __init__(self):
+        """Initialize the adapter."""
+        self.provider_adapter: Optional[BaseLLMAdapter] = None
+
+    def get_function_name(
+        self, function_def: Union[Dict[str, Any], FunctionSchema, FlowsFunctionSchema]
```
kompfner, line 51: What's the rationale for supporting both `FunctionSchema` and `FlowsFunctionSchema`?

Isn't the function-related stuff in the adapter only for handling understanding functions as specified in node definitions? If so—is there a use-case for someone passing a pipecat `FunctionSchema` in a pipecat flows definition?

src/pipecat_flows/manager.py
```diff
@@ -575,52 +508,48 @@ async def set_node(self, node_id: str, node_config: NodeConfig) -> None:
             tools = []
             new_functions: Set[str] = set()
 
+            async def process_schema(schema):
```
kompfner, line 511: Nit: "process" is vague. `register_function_schema`?

src/pipecat_flows/manager.py
```diff
@@ -763,49 +692,47 @@ def _validate_node_config(self, node_id: str, config: NodeConfig) -> None:
...
-                )
-            )
+            # Check for handler, transition_to, and transition_callback depending on format
+            if isinstance(func, FlowsFunctionSchema):
+                # For FlowsFunctionSchema, we can access the fields directly
+                has_handler = func.handler is not None
+                has_transition_to = func.transition_to is not None
+                has_transition_callback = func.transition_callback is not None
+            else:
+                # For dictionary formats, use the provider-specific format checks
```
kompfner, line 712: this is much easier to read than before 🥳

kompfner (approved):
Really nice improvement! Some minor comments only.

examples/dynamic/insurance_gemini.py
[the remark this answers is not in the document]
markbackman: Makes sense. I'm putting them back inline for this file.

For restaurant_reservation.py, they're used multiple times, so I'll leave them as is.

src/pipecat_flows/adapters.py
markbackman: No, you're right. This should be removed. I had originally thought about supporting both, but that really doesn't make sense since FlowsFunctionSchema is the same and has its Flows-specific args as Optional.

src/pipecat_flows/manager.py
markbackman: Good idea!

PR #117, opened 2025-03-26
Fix handling of function definitions that specify `"parameters": None…

kompfner:
…` (a pattern shown in some of our demos—specifically, it seems, the ones that use Gemini and function calling)


markbackman (approved):
Nice! Is there anything to update in the demos as a result of this change?

PR #128, opened 2025-05-14
Support waiting until the user speaks to run LLM inference in a node,…

kompfner:
… via a new optional `NodeConfig.respond_immediately` property

### DIFF
--- a/examples/runner.py
+++ b/examples/runner.py
@@ -8,7 +8,11 @@ import argparse
 import os
 
 import aiohttp
-from pipecat.transports.services.helpers.daily_rest import DailyRESTHelper, DailyMeetingTokenParams, DailyMeetingTokenProperties
+from pipecat.transports.services.helpers.daily_rest import (
+    DailyMeetingTokenParams,
+    DailyMeetingTokenProperties,
+    DailyRESTHelper,
+)
 
 
 async def configure(aiohttp_session: aiohttp.ClientSession):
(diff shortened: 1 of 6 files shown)


kompfner:
Still doing additional testing, but wanted to get this open for review

examples/runner.py
kompfner, line 1: The changes in this file were made by `ruff`, after running:
```
ruff format
ruff check --select I --fix
```

src/pipecat_flows/actions.py
kompfner, line 1: The changes in this file were made by `ruff`, after running:
```
ruff format
ruff check --select I --fix
```

src/pipecat_flows/manager.py
```diff
@@ -558,7 +558,7 @@ async def register_function_schema(schema):
             self.current_functions = new_functions
 
             # Trigger completion with new context
-            if self._context_aggregator:
+            if self._context_aggregator and node_config.get("respond_immediately", True):
```
kompfner, line 561: Here's the only actual Pipecat Flows logic change in this PR

markbackman (approved):
LGTM! 🚀

kompfner:
> Still doing additional testing, but wanted to get this open for review

The new flag seems to work well for for the initial node, which is the important part. 

I also tested setting it on subsequent nodes in the flow, and it seems to work (though obviously in this example flow—and in most?—it wouldn't make sense to set it on anything but the first node).

kompfner:
UPDATE: reverting this change. Setting `respond_immediately: False` on a node will cause any `post_actions` on it to run immediately rather than when the bot has finished responding (which is when I'd expect it to run).

PR #122, opened 2025-04-08
Add support for Amazon Bedrock LLMs

adithyaxx:
Summary
This PR adds a BedrockAdapter to interact with BedrockLLMService, enabling support for Amazon Bedrock LLMs within Pipecat Flows.

Depends On:
https://github.com/pipecat-ai/pipecat/pull/1527

PR #129, opened 2025-05-14
Update AWSBedrockAdapter

markbackman:
[This PR](https://github.com/pipecat-ai/pipecat-flows/pull/122) was submitted before the AWS Bedrock change was merged in Pipecat. This PR resolves some issues and aligns the naming between the Pipecat service (`AWSBedrockLLMService`) and the adapter (`AWSBedrockAdapter`).

Also:
- Adding two examples that use `AWSBedrockLLMService`.
- Fixing an issue with a Gemini example
- Adding tests for the new adapter
- Adding a changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added support for `AWSBedrockLLMService` by adding an `AWSBedrockAdapter`.
+
 ### Changed
 
 - Updated to use `FunctionCallParams` as args for the function handler.
@@ -14,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated imports to use the new .stt, .llm, and .tts paths.
 
 ### Other
+
+- Added AWS Bedrock examples for insurance and patient_intake.
 
 - Updated examples to `audio_in_enabled=True` and remove `vad_enabled` and
   `vad_audio_passthrough` to align with the latest Pipecat `TransportParams`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Bumped the minimum required `pipecat-ai` version to 0.0.67 to align with AWS
+  Bedrock additions in Pipecat. This also adds support for `FunctionCallParams`
+  which were added in 0.0.66.
+
 - Updated to use `FunctionCallParams` as args for the function handler.
 
 - Updated imports to use the new .stt, .llm, and .tts paths.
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,17 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+
+- Added support for `AWSBedrockLLMService` by adding an `AWSBedrockAdapter`.
+
 ### Changed
 
-- Added `respond_immediately` to `NodeConfig`. Setting it to `False` has the effect of making the
-  bot wait, after the node is activated, for the user to speak before responding.
+- Added `respond_immediately` to `NodeConfig`. Setting it to `False` has the
```
kompfner, line 16: Here we're accidentally adding back the changelog entry that was reverted

examples/dynamic/insurance_aws_bedrock.py
```diff
@@ -0,0 +1,408 @@
+#
+# Copyright (c) 2024, Daily
+#
+# SPDX-License-Identifier: BSD 2-Clause License
+#
+"""Insurance Quote Example using Pipecat Dynamic Flows with Google Gemini.
```
kompfner, line 6: Copy-paste issue: this comment refers to Google Gemini several times

src/pipecat_flows/adapters.py
```diff
@@ -27,7 +27,16 @@
 from pipecat.adapters.services.anthropic_adapter import AnthropicLLMAdapter
 from pipecat.adapters.services.gemini_adapter import GeminiLLMAdapter
 from pipecat.adapters.services.open_ai_adapter import OpenAILLMAdapter
-from pipecat.adapters.services.bedrock_adapter import BedrockLLMAdapter
+
+try:
```
kompfner, line 31: why does this not match the pattern of our importing any of the other adapters?

kompfner: afaik it should be safe to import these adapters. they're not gated behind optional dependencies

kompfner (approved):
A few comments. Otherwise, LGTM!

src/pipecat_flows/adapters.py
kompfner: ohhh is this because you don't want to have to bump the version of pipecat that pipecat flows depends on? what's your philosophy on version lag between pipecat flows and pipecat?

kompfner: left a comment to the same effect [here](https://github.com/pipecat-ai/pipecat-flows/pull/131), but: are we sure we want to be in the game of ensuring that pipecat flows works with older versions of pipecat? my hunch is it'd be fine to just bump the required version of pipecat in our .toml file

CHANGELOG.md
markbackman: Thanks!

examples/dynamic/insurance_aws_bedrock.py
markbackman: Good catch! 🕵️

src/pipecat_flows/adapters.py
markbackman: Reverting this change based on our discussion. I'm bumping the minimum Pipecat version instead to 0.0.67.

PR #131, opened 2025-05-14
Maintain backwards compatibility for verbose function call params

markbackman:
I'm open to cleaner ways to support this. The nice thing about this approach is:
a) it works
b) it'll be easy to remove when the time comes

### DIFF
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -36,8 +36,36 @@ from pipecat.frames.frames import (
     LLMSetToolsFrame,
 )
 from pipecat.pipeline.task import PipelineTask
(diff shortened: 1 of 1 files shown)


kompfner:
Are we sure we want to be in the game of ensuring that pipecat flows works with older versions of pipecat? We already do a reasonable-effort attempt at API stability between versions of pipecat. My personal hunch is that:
- It's reasonable to say "the latest version of pipecat flows requires the latest version of pipecat"
- It could be end up being _very_ hairy to try to have a support matrix between pipecat flows and several older versions of pipecat

markbackman:
Agreed. Cross-library, cross-version compatibility will get very messy. I'm going to bump the minimum version instead.

PR #133, opened 2025-05-15
Support waiting until user speaks attempt 2

kompfner:
This PR is the same as the now-reverted https://github.com/pipecat-ai/pipecat-flows/pull/128, but with [an additional change](https://github.com/pipecat-ai/pipecat-flows/pull/133/commits/29a403199d4fcee57f5315a69b5974ad2dc3280e) to ensure that `post_actions` (which today really behave like, and probably should be named, `post_first_response_actions`) run at the appropriate time: after the bot's first response in the node.

Here's the accompanying docs PR: https://github.com/pipecat-ai/docs/pull/236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Added `respond_immediately` to `NodeConfig`. Setting it to `False` has the effect of making the
+  bot wait, after the node is activated, for the user to speak before responding.
+
 - Bumped the minimum required `pipecat-ai` version to 0.0.67 to align with AWS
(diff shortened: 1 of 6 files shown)


kompfner:
Note to self: open docs PR describing this new functionality, too

kompfner:
> One thing I'm wondering: would it be clearer if the option name specifically called out the fact that it waits for the user to speak first? Like something along the lines of `await_user_speak_first`, `user_speaks_first`, or `respond_only_after_user`?

Not a big fan of `await_user_speak_first` because it doesn't make clear that it's only the LLM _response_ that's awaiting the user speech, not doing other node-entering stuff like setting the LLM context.

markbackman (approved):
I think `respond_immediately` makes sense.

Assuming changes are made, this looks great!

PR #137, opened 2025-05-19
Fix the warm transfer demo's hold music player, which stopped working…

kompfner:
… with daily-python 0.18.0.

The "fix" is actually just a workaround: the hold music player is now based on daily-python's `AsyncSendWavApp` example rather than `SendWaveApp`, which seems to have stopped working with daily-python 0.18.0.

### DIFF
--- a/examples/assets/hold_music/hold_music.py
+++ b/examples/assets/hold_music/hold_music.py
@@ -1,12 +1,13 @@
 #
 # This demo will join a Daily meeting and send the audio from a WAV file into
-# the meeting.
+# the meeting. It uses the asyncio library.
 #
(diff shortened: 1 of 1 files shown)


kompfner:
@aconchillo this is totally a non-fix—happy to look into the issue of why `SendWaveApp` stopped working with daily-python 0.18.0.

kompfner:
> @aconchillo this is totally a non-fix—happy to look into the issue of why `SendWaveApp` stopped working with daily-python 0.18.0 tomorrow. But in the meantime let's apply the workaround in this PR until that underlying issue is fixed.

And [here's the fix for the underlying issue](https://github.com/daily-co/daily-x/pull/2078). We should still probably merge this PR because it'll be a minute until the latest daily-python makes it way to pipecat flows.

Issue #132, opened 2025-05-15
Simplify Node Creation Syntax, Especially for Functions


badbye:
Currently, the syntax for creating nodes is somewhat complex, especially when it comes to the function call-related parts. One has to manually list the parameters required by each function in JSON.
The way to create a livekit agent is more concise and intuitive: https://github.com/livekit/agents/?tab=readme-ov-file#multi-agent-handoff

I think a good design that enables developers to get started quickly is conducive to the promotion of the project. May I ask if there are any plans to update the creation syntax at present?

markbackman:
Just so you're aware, Pipecat also offers very simple ways to do basic things.

For example, here's a basic voice bot that delivers equivalent functionality to what you linked to from Livekit:
```
#
# Copyright (c) 2024–2025, Daily
#
# SPDX-License-Identifier: BSD 2-Clause License
#

import argparse
import os

from dotenv import load_dotenv
from loguru import logger

from pipecat.audio.vad.silero import SileroVADAnalyzer
from pipecat.pipeline.pipeline import Pipeline
from pipecat.pipeline.runner import PipelineRunner
from pipecat.pipeline.task import PipelineParams, PipelineTask
from pipecat.processors.aggregators.openai_llm_context import OpenAILLMContext
from pipecat.services.cartesia.tts import CartesiaTTSService
from pipecat.services.deepgram.stt import DeepgramSTTService
from pipecat.services.openai.llm import OpenAILLMService
from pipecat.transports.base_transport import TransportParams
from pipecat.transports.network.small_webrtc import SmallWebRTCTransport
from pipecat.transports.network.webrtc_connection import SmallWebRTCConnection

load_dotenv(override=True)


async def run_bot(webrtc_connection: SmallWebRTCConnection, _: argparse.Namespace):
    logger.info(f"Starting bot")

    transport = SmallWebRTCTransport(
        webrtc_connection=webrtc_connection,
        params=TransportParams(
            audio_in_enabled=True,
            audio_out_enabled=True,
            vad_analyzer=SileroVADAnalyzer(),
        ),
    )

    stt = DeepgramSTTService(api_key=os.getenv("DEEPGRAM_API_KEY"))

    tts = CartesiaTTSService(
        api_key=os.getenv("CARTESIA_API_KEY"),
        voice_id="71a7ad14-091c-4e8e-a314-022ece01c121",  # British Reading Lady
    )

[...]

PR #141, opened 2025-05-22
Function simplification exploration

kompfner:
## Overview

This PR introduces two concepts/features related to functions (tools):

1. **"Consolidated" functions**, which are responsible for _both_ data processing (i.e. the work that `handler`s are responsible for) as well as specifying the next node to transition to (i.e. the work that always previously fell to `transition_to` and `transition_callback`).
2. **"Direct" functions**, which are Python functions that the user can pass in _directly_ to `functions`, in lieu of function declaration objects or `FlowsFunctionSchema`s.

**Direct functions are expected to always be consolidated functions.**

This work was inspired by https://github.com/pipecat-ai/pipecat-flows/issues/132.

## Consolidated functions

When specifying a consolidated function, you don't need to provide either a `transition_to` or `transition_callback`. You can also avoid having to read stashed-away data values from a `FlowResult` when deciding how to transition, since the transition logic lives in the same function as the data processing.

### Usage

```python
async def check_availability(args: FlowArgs) -> tuple[TimeResult, NodeConfig]:
    """Check reservation availability and return result."""
    time = args["time"]
    party_size = args["party_size"]

    # Check availability with mock API
    is_available, alternative_times = await reservation_system.check_availability(party_size, time)

    result = TimeResult(
        status="success", time=time, available=is_available, alternative_times=alternative_times
    )

    if is_available:
        logger.debug("Time is available, transitioning to confirmation node")
        next_node = create_confirmation_node()
    else:
        logger.debug(f"Time not available, storing alternatives: {result['alternative_times']}")
        next_node = create_no_availability_node(alternative_times)

    return result, next_node

availability_schema = FlowsFunctionSchema(
    name="check_availability",
    description="Check availability for requested time",
    properties={
        "time": {
            "type": "string",
            "pattern": "^([5-9]|10):00 PM$",  # Matches "5:00 PM" through "10:00 PM"
            "description": "Reservation time (e.g., '6:00 PM')",
        },
        "party_size": {"type": "integer"},
    },
    required=["time", "party_size"],
    handler=check_availability,
    transition_callback=handle_availability_check,
)

def create_time_selection_node() -> NodeConfig:
    """Create node for time selection and availability check."""
    logger.debug("Creating time selection node")
    return {
        "task_messages": [
            {
                "role": "system",
                "content": "Ask what time they'd like to dine. Restaurant is open 5 PM to 10 PM.",
            }
        ],
        "functions": [availability_schema],
    }
```

## Direct functions

When using direct functions, you don't need to explicitly specify `name`, `description`, `properties`, or `required`; those values are all automatically parsed from your direct function's signature and docstring. This obviates the need to define function declaration objects or `FlowsFunctionSchema`s

### Usage

```python
async def check_availability(
    flow_manager: FlowManager, time: str, party_size: int
) -> tuple[TimeResult, NodeConfig]:
    """
    Check availability for requested time.

    Args:
        time (str): Requested reservation time in "HH:MM AM/PM" format. Must be between 5 PM and 10 PM.
        party_size (int): Number of people in the party.
    """
    # Check availability with mock API
    is_available, alternative_times = await reservation_system.check_availability(party_size, time)

    # Result: availability status and alternative times, if any
    result = TimeResult(
        status="success", time=time, available=is_available, alternative_times=alternative_times
    )

    # Next node: confirmation or no availability
    if is_available:
        next_node = create_confirmation_node()
    else:
        next_node = create_no_availability_node(alternative_times)

    return result, next_node

def create_time_selection_node() -> NodeConfig:
    """Create node for time selection and availability check."""
    logger.debug("Creating time selection node")
    return {
        "task_messages": [
            {
                "role": "system",
                "content": "Ask what time they'd like to dine. Restaurant is open 5 PM to 10 PM.",
            }
        ],
        "functions": [check_availability],
    }
```

There are some tradeoffs to using direct functions:
* You have a little less control over what appears in "description": for example, if you wanted to put more information in your docstring than you want to expose to the LLM when registering the tool, there's no way to do that.
* You have a little less control over each function property's definition (its JSON schema), as the properties are automatically read from the Python function. For example, there's no ability to specify strict explicit constraints like min and max values—you'd have to just put them in the arg description in the doc string.
* You don't have control over the printed name of the next node (yet...I might address that in an upcoming change in this PR)
* Overall, there's just a lot more "magic": you have to keep in mind that your docstring has some significance to the LLM, for example.


src/pipecat_flows/types.py
```diff
@@ -231,8 +231,13 @@ class FlowsFunctionSchema:
         properties: Dictionary defining properties types and descriptions
         required: List of required parameters
         handler: Function handler to process the function call
-        transition_to: Target node to transition to after function execution
-        transition_callback: Callback function for dynamic transitions
+        transition_to: Target node to transition to after function execution (deprecated)
+        transition_callback: Callback function for dynamic transitions (deprecated)
+
+    Deprecated:
+        0.0.18: `transition_to` and `transition_callback` are deprecated and will be removed in a
```
kompfner, line 238: Exact version TBD here. Note to self: make sure to update this, if needed, before merging this PR.

kompfner:
Went ahead and ended up already implementing in this PR some of the more "forward-looking" changes we discussed, @markbackman. Not doing them in one fell swoop meant introducing some temporary API awkwardness which I figured we could just leapfrog. See latest commit messages for rationale.

Marking this PR as ready for review, even though I'm still working through the examples to bring them up to the new recommended patterns and avoid deprecation warnings.

examples/dynamic/insurance_anthropic.py
```diff
@@ -88,22 +88,37 @@ class CoverageUpdateResult(FlowResult, InsuranceQuote):
...
-    """Process age collection."""
+async def collect_age(
+    args: FlowArgs, flow_manager: FlowManager
+) -> tuple[AgeCollectionResult, NodeConfig]:
+    """Process age collection and return next node name."""
     age = args["age"]
     logger.debug(f"collect_age handler executing with age: {age}")
-    return AgeCollectionResult(age=age)
 
+    flow_manager.state["age"] = age
```
kompfner, line 98: Went ahead and updated all the insurance examples to store age in `flow_manager.state` and read it back out in `collect_marital_status()`, which _seemed_ to be what was originally intended in this demo, but we had a limitation a while ago where function handlers didn't get `flow_manager`s passed to them.

examples/dynamic/insurance_anthropic.py
```diff
@@ -116,15 +131,17 @@ async def calculate_quote(args: FlowArgs) -> QuoteCalculationResult:
...
         "coverage_amount": 250000,
         "deductible": 1000,
     }
+    next_node = create_quote_results_node(result)
+    return result, next_node
 
 
-async def update_coverage(args: FlowArgs) -> CoverageUpdateResult:
-    """Update coverage options and recalculate premium."""
+async def update_coverage(args: FlowArgs) -> tuple[CoverageUpdateResult, NodeConfig]:
```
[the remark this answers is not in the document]
kompfner: (Not relevant to this PR, just noticed this during testing)

kompfner:
Note to self: the changes in this PR will require docs site changes.

src/pipecat_flows/types.py
```diff
@@ -208,6 +268,181 @@ def to_function_schema(self) -> FunctionSchema:
         )
 
 
+class FlowsDirectFunction:
```
markbackman, line 271: Maybe to keep the types.py file focused on type definitions, the core logic could be moved elsewhere?

(Though, if this will ultimately live in Pipecat, maybe this is OK here since this will be removed in the future.)

markbackman:
In the manager.py, there's this code which has been deprecated for around 4 months (lines 365-374):
```
                    elif transition_callback:  # Dynamic flow
                        logger.debug(f"Dynamic transition for: {name}")
                        # Check callback signature
                        sig = inspect.signature(transition_callback)
                        if len(sig.parameters) == 2:
                            # Old style: (args, flow_manager)
                            await transition_callback(args, self)
                        else:
                            # New style: (args, result, flow_manager)
                            await transition_callback(args, result, self)
```

Maybe we can deprecate the old signature and just use the new style.

src/pipecat_flows/types.py
```diff
@@ -208,6 +268,181 @@ def to_function_schema(self) -> FunctionSchema:
...
+                required.append(name)
+
+            # Add parameter description from docstring
+            for doc_param in docstring_params:
+                if doc_param.arg_name == name:
+                    properties[name]["description"] = doc_param.description or ""
+
+        return properties, required
+
+    def _typehint_to_jsonschema(self, type_hint: Any) -> Dict[str, Any]:
```
markbackman, line 374: Magic ✨

src/pipecat_flows/types.py
```diff
@@ -208,6 +268,181 @@ def to_function_schema(self) -> FunctionSchema:
...
+    # TODO: maybe to better support things like enums, check if each type is a pydantic type and use its convert-to-jsonschema function
+    def _get_parameters_as_jsonschema(
+        self, func: Callable, docstring_params: List[docstring_parser.DocstringParam]
+    ) -> Tuple[Dict[str, Any], List[str]]:
+        """
+        Get function parameters as a dictionary of JSON schemas and a list of required parameters.
+        Ignore the last parameter, as it's expected to be the flow_manager.
+
+        Args:
+            func: Function to get parameters from
```
markbackman, line 333:
```suggestion
            func: Function to get parameters from
            docstring_params: List of parameters extracted from the function's docstring
```

src/pipecat_flows/types.py
```diff
@@ -208,6 +268,181 @@ def to_function_schema(self) -> FunctionSchema:
...
+        first_param_name = params[0][0]
+        if first_param_name != "flow_manager":
+            raise InvalidFunctionError(
+                f"Direct function {function.__name__} first parameter must be named 'flow_manager'"
+            )
+
+    async def invoke(
+        self, args: Mapping[str, Any], flow_manager: "FlowManager"
+    ) -> ConsolidatedFunctionResult:
+        return await self.function(flow_manager=flow_manager, **args)
```
markbackman, line 294: Maybe add docstring here:
```
        """Invoke the direct function with the given arguments and flow manager.

        Args:
            args: Dictionary of arguments to pass to the function
            flow_manager: Reference to the FlowManager instance
        Returns:
            ConsolidatedFunctionResult: Result of the function execution, which can include both a
                FlowResult and the next node to transition to.
        """
```

markbackman:
This looks really, really good!

I'm already looking forward to the future where we can remove deprecations. We've acquired a bunch a long the way, which is why I'm encouraging to remove anything that's ~3 months or older. Maybe the old function signature deprecations can go away which could simplify just a bit.

examples/dynamic/insurance_anthropic.py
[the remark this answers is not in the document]
markbackman: age and marital status are used to calculate the quote. I think this is OK to keep it simple, but you're free to make any changes you'd like!

src/pipecat_flows/manager.py
```diff
@@ -134,6 +143,9 @@ def __init__(
         self.current_functions: Set[str] = set()  # Track registered functions
         self.current_node: Optional[str] = None
 
+        self._showed_deprecation_warning_for_transition_fields = False
```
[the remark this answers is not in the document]
kompfner: Given that the deprecation warnings are relatively verbose and a codebase is likely to have various instances of deprecated usage, it felt a bit overwhelming to me to see many repeated instances of the same deprecation warning. This one-flag complexity feels low enough to be worth it, in my opinion.

src/pipecat_flows/types.py
```diff
@@ -250,14 +487,20 @@ class NodeConfig(NodeConfigRequired, total=False):
...
     role_messages: List[Dict[str, Any]]
-    functions: List[Union[Dict[str, Any], FlowsFunctionSchema]]
+    functions: List[Union[Dict[str, Any], FlowsFunctionSchema, DirectFunction]]
     pre_actions: List[ActionConfig]
     post_actions: List[ActionConfig]
     context_strategy: ContextStrategyConfig
     respond_immediately: bool
 
 
+def get_or_generate_node_name(node_config: NodeConfig) -> str:
```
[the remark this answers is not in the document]
kompfner: Yeah, the intent was to make it so that folks wouldn't _have_ to immediately start using the new `NodeConfig.name` field (so yes, backwards compatibility is the concern).

Though...we _could_ make it so that if you—the user—wanted to do anything one of the "new" ways, i.e.
- specify an `initial_node` in `initialize()`
- use a "consolidated" function that returns `(result, next_node)`
- call `set_node_from_config()`

...then we'd require you to specify a `name`. That way we could pave the way towards making creating a `NodeConfig` without a `name` deprecated usage.

I'm undecided whether that's worth doing.

src/pipecat_flows/manager.py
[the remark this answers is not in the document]
markbackman: Good point!

src/pipecat_flows/types.py
kompfner: > focused on type definitions

types.py contains things like `FlowsFunctionSchema`, which does include some validation and conversion logic. `FlowsDirectFunction` is essentially analogous, but happens to require a lot more lines of code to do the same thing.

To me, it makes more sense to make grouping decisions by similar functionality instead of by lines of code.

But you're totally right, this will all be moot once we port most of this logic "down" to the Pipecat layer.

kompfner:
> Maybe we can deprecate the old signature and just use the new style.

Agree. I'm adding this to a list of deprecated things to finally sunset. Let's do this in a separate PR.

markbackman (approved):
Looks really great! What a nice simplification 👏 

I've tested a bunch of the examples and see the expected behavior. Ship it! 🚢

PR #142, opened 2025-06-03
Pre- and post-actions fixes

kompfner:
Highly recommend looking at individual commit messages for explanation.

### DIFF
--- a/src/pipecat_flows/actions.py
+++ b/src/pipecat_flows/actions.py
@@ -190,6 +190,12 @@ class ActionManager:
                 # Record the type of the action we just executed
                 previous_action_type = action_type
                 logger.debug(f"Successfully executed action: {action_type}")
+
+                # If action was end_conversation, break
+                # (If we didn't, we could end up waiting for the next actions to finish, and...they
+                # never would)
+                if action_type == "end_conversation":
+                    break
             except Exception as e:
                 # Undo any increment of ongoing actions count that happened during this action
                 if self._ongoing_actions_count > ongoing_actions_count:
@@ -246,14 +252,6 @@ class ActionManager:
             # - custom action: wait (we don't know what it will do)
             if upcoming_action_type not in ["tts_say", "end_conversation", "function"]:
                 needs_wait = True  # None or custom action
-        elif previous_action_type == "end_conversation":
-            # "end_conversation" enqueues an EndFrame, which has an effect at the end of the
-            # pipeline.
-            # Although we *should* wait here (we don't want other actions to sneak ahead of
-            # "end_conversation"), we don't, because we can't really detect when it's done (the
-            # stopped pipeline never delivers the ActionFinishedFrame). If we waited, we'd end up
-            # waiting forever.
-            pass
         elif previous_action_type == "function":
             # "function" enqueues a FunctionActionFrame, which has an effect at the end of the
             # pipeline.
@@ -270,6 +268,8 @@ class ActionManager:
             #   tell us when the action is done. But let's hold off on doing that since we're
             #   de-emphasizing custom actions in favor of "function" actions, which should meet most
             #   needs.
+            # Note that it should not be possible for the previous action to be "end_conversation",
+            # since we stop processing actions after that one.
             pass
 
         if needs_wait:
--- a/src/pipecat_flows/actions.py
+++ b/src/pipecat_flows/actions.py
@@ -150,6 +150,7 @@ class ActionManager:
             if not handler:
                 raise ActionError(f"No handler registered for action type: {action_type}")
 
+            ongoing_actions_count = self._ongoing_actions_count
             try:
                 # Based on the type of the previous action and the one coming up, we can determine
                 # if we need to wait for ongoing actions to finish before proceeding with this next
@@ -191,6 +192,9 @@ class ActionManager:
                 previous_action_type = action_type
                 logger.debug(f"Successfully executed action: {action_type}")
             except Exception as e:
+                # Undo any increment of ongoing actions count that happened during this action
+                if self._ongoing_actions_count > ongoing_actions_count:
+                    self._decrement_ongoing_actions_count()  # Assumption: on increment per action
                 raise ActionError(f"Failed to execute action {action_type}: {str(e)}") from e
 
         # Based on the type of the last action, we may need to wait for ongoing actions to finish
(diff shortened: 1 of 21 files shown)


src/pipecat_flows/actions.py
```diff
@@ -183,6 +211,55 @@ async def _execute_deferred_post_actions(self) -> None:
...
+            # - "function": no need to wait (effect happens at the end of the pipeline)
+            # - custom action: wait (we don't know what it will do)
+            if upcoming_action_type == "tts_say":
+                needs_wait = True
+            elif upcoming_action_type not in ["end_conversation", "function"]:
+                needs_wait = True
+        else:
+            # Either previous action was:
+            # - None (the upcoming action is the first one), so there's nothing to wait for.
+            # - A fully custom action, where we don't know if it enqueued a ActionFinishedFrame that
```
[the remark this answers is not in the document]
kompfner: It should essentially do the 2nd half of `_queue_action_frame()`:
- increment `self._ongoing_actions_count`
- if `self._ongoing_actions_count` is 1, `self._ongoing_actions_finished_event.clear()`

kompfner:
Up next: think about some renames and maybe adding a new type of lifecycle action hook:
- `pre_actions` -> `node_entry_actions`
- `post_actions` -> `post_first_response_actions`
- `node_exit_actions` (new)

kompfner:
> Up next: think about some renames and maybe adding a new type of lifecycle action hook:
> 
> * `pre_actions` -> `node_entry_actions`
> * `post_actions` -> `post_first_response_actions`
> * `node_exit_actions` (new)

This should come in a separate PR.

src/pipecat_flows/actions.py
```diff
@@ -72,25 +79,35 @@ def __init__(self, task: PipelineTask, flow_manager: "FlowManager", tts=None):
...
         @task.event_handler("on_frame_reached_downstream")
         async def on_frame_reached_downstream(task, frame):
             if isinstance(frame, FunctionActionFrame):
+                # Run function action
                 await frame.function(frame.action, flow_manager)
-                self.function_finished_event.set()
+                self._decrement_ongoing_actions_count()
             elif isinstance(frame, BotStoppedSpeakingFrame):
-                await self._execute_deferred_post_actions()
+                # Execute deferred post-actions if the bot's turn is over.
```
markbackman, line 103: Just mentioning since it's possible: during a bot's turn it is possible for the speech to be split up across multiple BotStartedSpeakingFrame/BotStoppedSpeakingFrame pairs. This is most likely with an HTTP based TTS service or a very slow cadence of speech from the model.

Generally, I think we _should_ be able to rely on a bot speaking contiguously, so if we hit issues, we may have to adjust the bot's VAD setting in Pipecat to ensure speech is, in fact, contiguous.

src/pipecat_flows/actions.py
```diff
@@ -131,7 +149,15 @@ async def execute_actions(self, actions: Optional[List[ActionConfig]]) -> None:
...
 
+            ongoing_actions_count = self._ongoing_actions_count
             try:
+                # Based on the type of the previous action and the one coming up, we can determine
+                # if we need to wait for ongoing actions to finish before proceeding with this next
+                # one
+                await self._maybe_wait_for_ongoing_actions_to_finish(
+                    previous_action_type, action_type
+                )
+
```
markbackman, line 160: Maybe we can remove the multiple signatures below. Could be a separate PR, but it would be nice to remove some old deprecations.

src/pipecat_flows/actions.py
```diff
@@ -183,26 +225,78 @@ async def _execute_deferred_post_actions(self) -> None:
...
+        needs_wait = False
+        if previous_action_type == "tts_say":
+            # "tts_say" enqueues a TTSSpeakFrame, which has an effect when it hits the TTS node in
+            # the pipeline.
+            # As long as the upcoming action enqueues a frame with an effect at the same point or
+            # later in the pipeline, we don't need to wait.
+            # If the upcoming action is:
+            # - "tts_say": no need to wait (effect happens at the same point)
+            # - "end_conversation": no need to wait (effect happens at the end of the pipeline)
+            # - "function": no need to wait (effect happens at the end of the pipeline)
```
markbackman, line 249: What do you mean "end of the pipeline"?

src/pipecat_flows/actions.py
```diff
@@ -183,26 +225,78 @@ async def _execute_deferred_post_actions(self) -> None:
...
+            # - custom action: wait (we don't know what it will do)
+            if upcoming_action_type not in ["tts_say", "end_conversation", "function"]:
+                needs_wait = True  # None or custom action
+        elif previous_action_type == "function":
+            # "function" enqueues a FunctionActionFrame, which has an effect at the end of the
+            # pipeline.
+            # Functions can take some time to execute (and don't hold up the pipeline as they're
+            # doing so), so we need to wait for them to finish before proceeding with the next
+            # action or moving on from the current set of actions.
+            needs_wait = True
```
markbackman, line 261: And we're conservative about custom_actions (i.e. we'll wait) so that we ensure ordering is protected, is that right?

markbackman (approved):
Ok, this makes sense and is more careful about how actions are processed. This is a nice improvement, which is evident by the fact that TTSSpeakFrame now works in the correct order.

I think it makes sense to proceed with this change.

src/pipecat_flows/actions.py
kompfner: Agree, that'd be nice

src/pipecat_flows/actions.py
kompfner: Well...sort of.

There are two kinds of user-defined actions now:
* Function actions (those with `type="function"`)
* Fully custom actions (with with a user-defined `type`)

We always wait for function actions to say that they're done before we move onto executing the next action.

We *don't* wait after a fully custom action, because we can't know when it's done with its work. It might just log something and be done immediately. Or it might queue a frame into the pipeline (our examples show it queueing a frame). We simply don't know, and there's no way today for the user to tell us what the end point of that action is (we'd need some new API affordances). This is exactly the situation we were in prior to this PR.

kompfner:
> Note to self: the deprecation of the `tts` `FlowManager` parameter will require docs site changes.

Huh. Actually, looks like there's already no mention of the `tts` parameter in the docs site. So...hooray!

PR #144, opened 2025-06-11
Fix `TestContextStrategies.test_provider_specific_summary_formatting`…

kompfner:
… unit test, which is broken when used with Pipecat versions 0.0.68 and above


markbackman (approved):
👍

PR #147, opened 2025-06-13
Fix: Handle run_in_parallel=False, simplify pending function call tra…

markbackman:
…cking

#145 but on top of the `pk/function-simplification-exploration` branch. This now relies on Pipecat to do the pending function call tracking and removes the counter and tracking logic from Flows, which is a nice simplification.

DO NOT MERGE YET.

Requires an updated pipecat-ai version due to: https://github.com/pipecat-ai/pipecat/pull/2006.

### DIFF
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -588,6 +588,10 @@ In all of these cases, you can provide a `name` in your new node's config for de
             raise FlowTransitionError(f"{self.__class__.__name__} must be initialized first")
 
         try:
+            # Clear any pending transition state when starting a new node
+            # This ensures clean state regardless of how we arrived here:
+            # - Normal transition flow (already cleared in _check_and_execute_transition)
+            # - Direct calls to set_node/set_node_from_config
             self._pending_transition = None
 
             self._validate_node_config(node_id, node_config)


markbackman:
@kompfner this is ready for review.

src/pipecat_flows/manager.py
```diff
@@ -439,47 +362,97 @@ async def transition_func(params: FunctionCallParams) -> None:
...
                     is_transition_only_function = True
+
                 logger.debug(
                     f"{'Transition-only function called for' if is_transition_only_function else 'Function handler completed for'} {name}"
                 )
 
-                # For edge functions, prevent LLM completion until transition (run_llm=False)
-                # For node functions, allow immediate completion (run_llm=True)
+                # Determine if this is an edge function
                 has_explicit_transition = bool(transition_to) or bool(transition_callback)
```
kompfner, line 371: Super nit: the only reason we were storing `has_explicit_transition` before was to help determine how to invoke `on_context_updated_edge()` (though, looking back on it, I don't think it was necessary). If we don't need this var anymore, we can just get rid of it and just do:

```
is_edge_function = bool(next_node) or bool(transition_to) or bool(transition_callback)
```

Otherwise the reader might think it's significant.

markbackman: Thanks for catching this 🙇

src/pipecat_flows/manager.py
```diff
@@ -614,6 +587,8 @@ async def _set_node(self, node_id: str, node_config: NodeConfig) -> None:
             raise FlowTransitionError(f"{self.__class__.__name__} must be initialized first")
 
         try:
+            self._pending_transition = None
```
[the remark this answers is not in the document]
markbackman: That's right. I think that makes sense, as an explicitly called function should take precedent. I'll add a comment.

PR #151, opened 2025-06-21
Fix handler handling of class and instance methods

### LINKED ISSUE #150: Function calls doesn't work when the handler is a method on an object
https://github.com/pipecat-ai/pipecat-flows/blob/b022068222b989ac38a91d0dd43c46373633284a/src/pipecat_flows/manager.py#L286

When using methods, manager is trying to subtract self from the prototype's parameter count. But it looks like self is already subtracted so it ends up calling the method with too few parameters.

trycatchal:
fixes #150

### DIFF
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -279,14 +279,8 @@ class FlowManager:
         # Get the function signature
         sig = inspect.signature(handler)
 
-        # Check if handler is a method (has self parameter)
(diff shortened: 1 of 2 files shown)


kompfner (approved):
Nice catch! And thanks for the great accompanying unit tests!

I wondered if `sig.parameters` counting `self` was maybe a Python 3.10 behavior that the code was originally accounting for (Python 3.10 being our lowest supported version), but...nope, doesn't seem to be.

PR #153, opened 2025-06-24
Refactor "direct" functions, now that function metadata parsing has b…

kompfner:
…een implemented in core Pipecat

---

Depends on https://github.com/pipecat-ai/pipecat/pull/2051 landing and being published first.

---

Note the (possibly confusing to the reviewer) name changes here:

- `DirectFunction` -> `FlowsDirectFunction`
- `FlowsDirectFunction` -> `FlowsDirectFunctionWrapper`


pyproject.toml
```diff
@@ -19,7 +19,7 @@ classifiers = [
     "Topic :: Multimedia :: Sound/Audio",
 ]
 dependencies = [
-    "pipecat-ai>=0.0.67",
+    "pipecat-ai>=0.0.73",
```
kompfner, line 22: Note: this is just a guess as to the version of pipecat-ai that will contain `BaseDirectFunctionWrapper`. I'll make sure to update this to the appropriate version if it seems like it'll differ from the one named here.

kompfner: Bumped the version to 0.0.74 now

PR #156, opened 2025-06-27
fix: update generate_summary in the Google adapter to use the google-…

markbackman:
…genai APIs

In Pipecat ~0.0.68,`pipecat-ai` updated to use the `google-genai` package. `pipecat-ai-flows` depends on the packages installed in `pipecat-ai`. This is an update from the now deprecated `google-generativeai` package to the new `google-genai` package.

Also, I took the opportunity to better construct the prompt—the `summary_prompt` is now part of the `system_instruction`.

### DIFF
--- a/src/pipecat_flows/adapters.py
+++ b/src/pipecat_flows/adapters.py
@@ -462,18 +462,30 @@ class GeminiAdapter(LLMAdapter):
     ) -> Optional[str]:
         """Generate summary using Google's API directly."""
         try:
-            # Format messages for Gemini
+            from google.genai.types import Content, GenerateContentConfig, Part
+
+            # Format conversation history as user message
             contents = [
-                {
-                    "role": "user",
-                    "parts": [{"text": (f"{summary_prompt}\n\nConversation history: {messages}")}],
-                }
+                Content(role="user", parts=[Part(text=f"Conversation history: {messages}")])
             ]
 
-            # Use non-streaming completion
-            response = await llm._client.generate_content_async(contents=contents, stream=False)
+            # Use summary_prompt as system instruction
+            generation_config = GenerateContentConfig(system_instruction=summary_prompt)
 
-            return response.text
+            # Use the new google-genai client's async method
+            response = await llm._client.aio.models.generate_content(
+                model=llm._model_name,
+                contents=contents,
+                config=generation_config,
+            )
+
+            # Extract text from response
+            if response.candidates and response.candidates[0].content:
+                for part in response.candidates[0].content.parts:
+                    if part.text:
+                        return part.text
+
+            return None
 
         except Exception as e:
             logger.error(f"Google summary generation failed: {e}", exc_info=True)


src/pipecat_flows/adapters.py
```diff
@@ -462,18 +462,30 @@ async def generate_summary(
     ) -> Optional[str]:
         """Generate summary using Google's API directly."""
         try:
-            # Format messages for Gemini
+            from google.genai.types import Content, GenerateContentConfig, Part
```
kompfner, line 465: To use these we'd need ""google-genai" in our pyproject.toml right?

kompfner: Or is it enough that these are transitive dependencies (I'm showing my python ignorance)

markbackman: No, to use the GeminiAdapter, you need to be using Pipecat with the `google` dependency installed, which installs the `google-genai` package. I think this is a safe assumption.

src/pipecat_flows/adapters.py
```diff
@@ -462,18 +462,30 @@ async def generate_summary(
...
             ]
 
-            # Use non-streaming completion
-            response = await llm._client.generate_content_async(contents=contents, stream=False)
+            # Use summary_prompt as system instruction
+            generation_config = GenerateContentConfig(system_instruction=summary_prompt)
 
-            return response.text
+            # Use the new google-genai client's async method
+            response = await llm._client.aio.models.generate_content(
```
kompfner, line 476: Oof it feels so fraught reaching into the LLM service and grabbing a private variable (`_client`). That's how we ended up with the breakage that necessitated this PR in the first place right?

kompfner: Well, this change doesn't make things _less_ fraught than before.

markbackman: No, that wasn't the root cause for this PR. The issue was that Flows depends on Pipecat bringing the LLM python package. Since that package changed, Flows needs to update to use the new API.

We could alternatively require optional dependencies in Flows, but that seems unnecessary since the package is already installed when using Pipecat. WDYT?

kompfner (approved):

CHANGELOG.md
```diff
@@ -128,16 +113,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   be removed in a future version. The `tts_say` action now pushes a
   `TTSSpeakFrame`.
 
+- Deprecated `transition_to` and `transition_callback` in favor of
```
markbackman, line 116: There were two Deprecation sections, which I consolidated.

Also, just some line break cleanup (keeping at 80, where it makes sense).

src/pipecat_flows/adapters.py
markbackman: I see what you mean. Yes, the `_client` in Pipecat changed, which it's not respecting the API. So, yes, in that case, this was the issue. Maybe some tech debt we should work around in the future.

PR #164, opened 2025-07-09
Add a quickstart README, add missing requirements.txt item

markbackman:
Adding a README for the quickstart.

Also, it was missing the `pipecat_ai_small_webrtc_prebuilt` package in requirements.txt.

### DIFF
--- a/examples/quickstart/README.md
+++ b/examples/quickstart/README.md
@@ -1,0 +1,53 @@
+# Pipecat Flows: Quickstart
+
+This quickstart example will help you set up your first Flow and explain to you the basics of using Flows.
+
+## Dependencies
+
+- Python 3.10 or higher
+- [Pipecat](https://github.com/pipecat-ai/pipecat?tab=readme-ov-file#-getting-started)
+
+## Installation
+
+1. Setup your virtual environment:
+
+```bash
+python3 -m venv venv
+source venv/bin/activate
+```
+
+2. Install Pipecat Flows:
+
+```bash
+pip install pipecat-ai-flows
+```
+
+3. Install the quickstart requirements:
+
+```bash
+pip install -r requirements.txt
+```
+
+4. Create a .env file with your API keys. For this demo, keys are required for [Cartesia](https://play.cartesia.ai/sign-up) and [Google Gemini](https://ai.google.dev/).
+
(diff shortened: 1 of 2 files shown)


examples/quickstart/README.md
```diff
@@ -0,0 +1,53 @@
...
+pip install pipecat-ai-flows
+```
+
+3. Install the quickstart requirements:
+
+```bash
+pip install -r requirements.txt
+```
+
+4. Create a .env file with your API keys. For this demo, keys are required for [Cartesia](https://play.cartesia.ai/sign-up) and [Google Gemini](https://ai.google.dev/).
```
kompfner, line 31: Do we want to have an `env.example` file in here, like we do in other example folders?

examples/quickstart/requirements.txt
```diff
@@ -1,4 +1,5 @@
 pipecat-ai[daily,silero,webrtc,websocket,cartesia,google]
```
kompfner, line 1: Do we want to not even include Daily? We don't mention it in the README at all...

examples/quickstart/README.md
markbackman: Maybe we should recommend updating the top-level .env file with these keys.

kompfner: Maybe! That's wasn't quite what I was getting at, though (sorry for being unclear!) I was more concerned about the reader being able to find out the names of the environment variables, hence my suggestion to point them towards env.example.

examples/quickstart/README.md
```diff
@@ -0,0 +1,53 @@
+# Pipecat Flows: Quickstart
```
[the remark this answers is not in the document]
kompfner: > That's probably something we should fix in the top-level README

Should we still address the issue of readers going through the top-level README hitting a stumbling block when they try to run the quickstart demo and getting an error because they didn't set up their .env file?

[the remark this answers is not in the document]
vipyne: rather than duplicate info, I suggest we link to quickstart readme in the top level README:
https://github.com/pipecat-ai/pipecat-flows/pull/167

kompfner (approved):

PR #172, opened 2025-07-21
Fix typo in task prompt of greeting node in movie_explorer_gemini.py

jacobggman:
Change the function names to be like movie_explorer_openai.py


markbackman (approved):
LGTM! Thanks for catching this.

PR #174, opened 2025-08-13
Pk/prototype llm failover attempt 4

kompfner:


kompfner:
> Hello @kompfner,
> 
> I would really like to help you move forward on this PR, however I'm having a lot of difficulty following your work. I can see that you've opened several PRs that depend on each other.
> 
> Could you let me know when and on which part I can come and help?
> 
> Maybe on the compatibility of your work with Anthropic and Mistral?

We've been working on the base Pipecat changes and are planning to release them with pipecat-ai 0.0.82 (the version after the one that we are planning on releasing today).

Good to know about your priorities re Anthropic and Mistral! Mistral support for the universal `LLMContext` should be relatively trivial (since the `MistralLLMService` is a subclass of the OpenAI one), and we can probably aim to include that in 0.0.82 as well.

Other than the OpenAI-based subclasses, the only other (non-speech-to-speech) LLM services that need migrating to the universal `LLMContext` pattern are Anthropic and AWS Bedrock. The migration work roughly encompasses:
- Supporting the `elif isinstance(frame, LLMContextFrame)` logic branch in `process_frame`
- Which entails supporting running the LLM inference starting with an `LLMContext` instead of an LLM-specific context (like `AnthropicLLMContext`)
- Which entails looking at any LLM-specific logic in the LLM-specific context and corresponding aggregators (e.g. `AnthropicLLMContext` and `AnthropicUser/AssistantContextAggregator` and confirming that the universal `LLMContext` and corresponding aggregators—`LLMUser/AssistantAggregator`—record a sufficiently expressive representation of context to work with the LLM. Any translation from the universal representation should then be ported into that LLM's adapt-from-universal-context method (i.e. `AnthropicLLMAdapter.get_llm_invocation_params()`)

src/pipecat_flows/adapters.py
```diff
@@ -84,17 +78,26 @@ def format_functions(
...
+    ) -> List[Dict[str, Any]] | ToolsSchema | NotGiven:
         """Format functions for provider-specific use.
 
+        The formatted functions/tools will be used in an LLMSetToolsFrame.
+
+        The NotGiven return type is only used by the subclass
+        UniversalLLMAdapter, as the universal LLMContext expects NOT_GIVEN when
+        there are no tools.
+
+        All other LLM-specific contexts expect either a list of dicts or a
```
kompfner, line 90: Note: I could've also invested in updating Pipecat's "legacy" aggregators to accept `NOT_GIVEN` as a signal for "there are no tools", and updated the Gemini adapter in this file to use `ToolsSchema`, but:
- I didn't want to make the "legacy" aggregators in Pipecat be aware of the "universal" `NOT_GIVEN` (even though under the hood it happens to be the same as OpenAI's `NOT_GIVEN`)
- I wanted to stay focused on the future and didn't want to muck around with or destabilize code paths that (hopefully) will be deprecated before too long.

examples/dynamic/insurance_gemini.py
```diff
@@ -343,8 +345,8 @@ async def main():
         tts = DeepgramTTSService(api_key=os.getenv("DEEPGRAM_API_KEY"), voice="aura-helios-en")
         llm = GoogleLLMService(api_key=os.getenv("GOOGLE_API_KEY"), model="gemini-2.0-flash-exp")
 
-        context = OpenAILLMContext()
-        context_aggregator = llm.create_context_aggregator(context)
+        context = LLMContext()
```
markbackman, line 348: Should we update all examples to follow this pattern?

src/pipecat_flows/adapters.py
markbackman: This makes sense.

examples/dynamic/insurance_gemini.py
kompfner: My thinking was that we should still have examples using the old pattern (which has not yet been deprecated) so we could continue exercising it until we deprecate it. WDYT?

Issue #178, opened 2025-08-22
Mistral function calling fails with "Unexpected role 'user' after role 'tool'" error in pipecat-flow


obigroup:
Hello @markbackman,

I've been testing Mistral with the latest version of pipecat 0.0.81 (which contains the Mistral implementation) in pipecat-flow, however I'm getting an error when the model calls a function:

```
openai.BadRequestError: Error code: 400 - {'object': 'error', 'message': "Unexpected role 'user' after role 'tool'", 'type': 'invalid_request_message_order', 'param': None, 'code': '3230'}
```

I can't seem to find where this might be coming from. Function calling works fine in `14w-function-calling-mistral.py`, so what does pipecat-flow do differently that could modify the message sent in:

```python
base_llm.py", line 209, in get_chat_completions
    chunks = await self._client.chat.completions.create(**params)
```

Could you give me a lead so I can make the necessary modifications and open a PR?

Thank you.

obigroup:
Here are more logs of the messages before the bug: 

```
{"role": "assistant", "tool_calls": [{"id": "UByW1YDuT", "function": {"name": "collect_order", "arguments": "{\"formated_order\": [{\"id\": 207499, \"quantity\": 1}]}"}, "type": "function"}]}, {"role": "tool", "content": "{\"status\": \"acknowledged\"}", "tool_call_id": "UByW1YDuT"}, {"role": "user", "content": "\n
```

PR #180, opened 2025-08-23
Add Workaround for Mistral Tool Message Handling

obigroup:
This code snippet is designed to manage the flow of messages in a conversation context. It initializes an empty list called `messages`. The purpose of the subsequent logic is to add an empty assistant message if the last message in the user's context has the role of "tool." This is a workaround specifically for Mistral, which requires an assistant message to follow any tool message.

The condition checks if there are any messages in the user's context. If there are, it verifies whether the last message's role is "tool." If both conditions are met, it appends a new message to the `messages` list with the role of "assistant" and a single space as the content. This space is crucial because an empty string does not function as intended.

This code has been added to `manager.py` to facilitate the transition from one node to another after a tool has been called by the language model (LLM).

### DIFF
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -903,6 +903,16 @@ In all of these cases, you can provide a `name` in your new node's config for de
         try:
             messages = []
 
+            # Add an empty assistant message if the last message has a role "tool".
+            # This is a workaround for Mistral, which requires an assistant message following a tool message.
+            if (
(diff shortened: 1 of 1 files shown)


obigroup:
ping @kompfner @markbackman

kompfner:
Thanks for this PR! Responded to the GH issue it's addressing [here](https://github.com/pipecat-ai/pipecat-flows/issues/178#issuecomment-3266475967). I suspect a different fix may be warranted.

PR #188, opened 2025-08-27
Fix warm transfer example AttributeError

jamsea:
## Description
Fixes the `AttributeError: 'FlowManager' object has no attribute 'state'` error in the warm transfer example.

## Changes
- Updated all references from `flow_manager.state` to `flow_manager._state` in `examples/dynamic/warm_transfer.py`
- The FlowManager class had changed its `state` attribute to `_state` (with underscore prefix) but the example code was still using the old attribute name

## Fixed Locations
- `start_hold_music` function (2 occurrences)
- `print_human_agent_join_url` function (1 occurrence)
- Main function setup (2 occurrences)  
- Cleanup function (1 occurrence)

## Testing
- ✅ Script now runs without errors
- ✅ Warm transfer functionality works correctly

## Fixes
Resolves the runtime error that prevented the warm transfer example from executing.

### DIFF
--- a/examples/dynamic/warm_transfer.py
+++ b/examples/dynamic/warm_transfer.py
@@ -165,8 +165,8 @@ async def mute_customer(action: dict, flow_manager: FlowManager):
 
 
 async def start_hold_music(action: dict, flow_manager: FlowManager):
-    hold_music_args = flow_manager.state["hold_music_args"]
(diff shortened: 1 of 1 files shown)


markbackman:
Oof! This is actually a Flows bug. I need to revert this. I was overzealous in a cleanup PR I did.

PR #195, opened 2025-09-04
Bump the pipecat dependency version to get Anthropic support for runt…

kompfner:
…ime LLM switching, and add Anthropic to the LLM switching example accordingly


kompfner:
> Note to self: just noticed a change is still needed to `AnthropicAdapter`. Stay tuned.

Done

PR #198, opened 2025-09-09
Bump the pipecat dependency version to get AWS Bedrock support for ru…

kompfner:
…ntime LLM switching, and add AWS Bedrock to the LLM switching example accordingly

**DO NOT MERGE until the changes in https://github.com/pipecat-ai/pipecat/pull/2619 have been published, presumably in Pipecat version 0.0.85.**


pyproject.toml
```diff
@@ -19,7 +19,7 @@ classifiers = [
     "Topic :: Multimedia :: Sound/Audio",
 ]
 dependencies = [
-    "pipecat-ai>=0.0.84",
+    "pipecat-ai>=0.0.85",
```
kompfner, line 22: Note: update this in case the corresponding Pipecat changes don't make it out in this version.

PR #202, opened 2025-09-17
Re-add uv.lock

markbackman:
Following the pattern for pipecat-ai. uv.lock only impacts developers. This version is generated with the pipecat-ai 0.0.85 minimum version.


kompfner (approved):
Agree, `uv.lock` should be there

PR #207, opened 2025-09-30
Fixed Deprecated Imports

lshaun:
Fixed Deprecated Imports from adapters.py by replacing

    ` module = sys.modules.get("pipecat.services.openai")`
    ` module = sys.modules.get("pipecat.services.openai.llm")`


markbackman (approved):
LGTM! Thanks for catching this.

PR #212, opened 2025-10-24
Add support for default tools in FlowManager initialization

Pranay-ai:
This PR adds support for default tools that persist across all nodes in the flow.
Key changes:

New default_tools parameter in FlowManager.__init__() (line 105) - accepts an optional list of FlowsFunctionSchema that will be available at every node
New instance variable _default_tools (line 153) - stores the default tools list, defaulting to an empty list if not provided
New tracking set _default_function_names (line 173) - tracks which function names belong to default tools (likely used elsewhere to ensure they're always registered alongside node-specific tools)

This allows callers to define tools once during initialization that will automatically be included at every node, rather than having to manually add them to each node's configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to **Pipecat Flows** will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added support for `global_functions` parameter in `FlowManager` initialization. Global functions are available at every node in a flow without needing to be specified in each node's configuration. Supports both `FlowsFunctionSchema` and `FlowsDirectFunction` types.
+
+### Changed
+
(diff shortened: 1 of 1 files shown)


markbackman:
Thanks for submitting this. Default tools that persist across nodes is a good idea!

@kompfner can you take a look when you get a moment?

kompfner:
Hey @Pranay-ai! Agreed, having a way to specify tools that apply to all nodes is a great idea!

I checked out your code to review it and ended up making a few changes. Will push them up to this branch in a minute. Here's a summary:

* Renamed `default_tools` to `global_functions`, for 2 reasons:
  * “global” communicates “cross-node” a bit more clearly than “default”, I think
  * in Pipecat Flows, “function” is more standard terminology than “tool”
* Added global function usage to the "main" food ordering examples, eliminating the need for a dedicated example
* Updated the example global function to make it a bit more natural/realistic: getting a delivery estimate
* Added support for `FlowsDirectFunction`s in addition to `FlowsFunctionSchema`s
* Simplified the implementation

kompfner (approved):

Pranay-ai (commented):
Great This Simplifies the Code and Workflow.

Pranay-ai:
> @Pranay-ai can you add a changelog entry for this? With that, I think we'll be ready to merge this down.

@markbackman  Great Will do that. Thanks

markbackman (approved):
Thanks again! 🙌

PR #217, opened 2025-11-19
add podcast example

Regaddi:
I changed the service selection to match what we have in the restaurant reservation example, including the LLM_PROVIDER switch.


markbackman (approved):
Looks great!

Issue #220, opened 2025-11-24
Add support for cancel_on_interruption paramater when registering functions.


DhirajDunani:
When registering functions directly via `llm.register_function()`, I can specify `cancel_on_interruption=False` to prevent certain functions from being cancelled when a user interrupts. 

However, when using pipecat-flows with FlowsFunctionSchema in node configurations, there's no way to specify this parameter.

PR #225, opened 2026-01-04
add cancel_on_interruption for FlowsFunctionSchema

smokyabdulrahman:
solves #220


smokyabdulrahman:
I will add tests if this solution is acceptable.

smokyabdulrahman:
I would appreciate your input here @markbackman 🙏

itissid:
Nice. That's really a small change. Hope it gets merged soon in the next pipecat release!

I've been using filler phrases to reduce perceived latency, but even so some LLM runs take many seconds longer, which can get cancelled when the user speaks.

I've extended STT mute as a workaround to not have functions cancelled, but that's not ideal since we lose what the user has said.

PR #232, opened 2026-02-02
Carry over "deactivated" versions of functions from one node to the next

### LINKED ISSUE #230: flows doesn't unregister_function() functions that are not in the new node
Hello friends.

When going from node A with functions F to node B with functions G, we should unregister_function() all functions in F-G.

I thought it's harmless to leave them in there, but these days we're seeing our LLM do a lot of hallucinated function calls that aren't in the tools list, and in those cases, we want the ~~unregistered function handler~~ catch-all handler to handle those instead of the should-not-be-available-~~function~~handler.

Thanks!

EDITED: Semantics

kompfner:
These changes were first made while exploring Gemini Live <-> Pipecat Flows interop. See CHANGELOG for a more detailed description.

Fixes: https://github.com/pipecat-ai/pipecat-flows/issues/230

Note that we're actually not (at least today) officially supporting Gemini Live in Pipecat Flows, but this feature is generally useful regardless.

Pipecat core companion PR, made during the exploration: https://github.com/pipecat-ai/pipecat/pull/3620

Why not already officially support Gemini Live in Pipecat Flows?

* The need to reconnect on context and tools updates makes Gemini Live not the most natural fit for Flows, performing notably slower than cascaded text-to-text pipelines during node transitions
* Because of this, the remaining investment to fully support Gemini Live with Flows (rethinking Flows node action timings for realtime models, in addition to ongoing long-term maintenance costs) doesn't—at least today—clear the necessary priority bar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed interrupted transition leaving flow permanently stuck when a user
   interruption cancelled a function call mid-execution (#234).
 
+- Added support for using Gemini Live (`GeminiLiveLLMService`) with Pipecat
+  Flows. Updated examples to support `LLM_PROVIDER=gemini_live`, accordingly.
+  Note that using Gemini Live requires that you use `LLMContext` and
+  `LLMContextAggregatorPair` rather than the deprecated `OpenAILLMContext` and
+  associated aggregators.
+
+### Changed
+
+- When transitioning from one node to the next, any functions in the previous
+  node that are absent in the new node now get "carried over", in that they're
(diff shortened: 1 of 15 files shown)


filipi87 (commented):
Hi @kompfner, is this PR still waiting for review ?

kompfner:
Hey @filipi87, @markbackman  - yeah, I had tabled this work for a while as more pressing work items came up (and we de-prioritized supporting Gemini Live in Flows). There is an open issue in this repo waiting for this work, though, so I plan to pick it back up and wrap it up over the coming days.

And yes, a review is needed (though keep in mind after the rebasing/conflict resolution there may be some changes).

kompfner:
After further discussion with the team, we're going to instead pursue a fix that lives in base Pipecat itself, as "deactivating" functions is not just a Pipecat Flows concern.

cc @markbackman

PR #247, opened 2026-04-01
Use "developer" role for non-user messages, bump pipecat-ai to >=0.0.108

kompfner:
pipecat-ai 0.0.108 adds support for "developer"-role messages in LLM context. Messages that are instructions from the application (not actual or simulated user utterances) should use the "developer" role.

Also update `generate_summary()` to use new the new `system_instruction` argument rather than putting the summary prompt in the context.


src/pipecat_flows/adapters.py
```diff
@@ -178,18 +178,14 @@ async def generate_summary(
 
             prompt_messages = [
                 {
-                    "role": "system",
-                    "content": summary_prompt,
-                },
-                {
-                    "role": "user",
+                    "role": "developer",
```
kompfner: While doing this, I noticed that in Pipecat we're still using "user" for the message containing the conversation history to summarize. We should update it to "developer".

markbackman (commented):
LGTM! Thanks for cleaning this up 🙏

markbackman (commented):
Changelog entry?

PR #249, opened 2026-04-02
pyproject.toml package and security updates

markbackman:
Update packages to align with Pipecat.

Update uv.lock with packages to resolves security vulnerabilities.

Also, this project isn't configured for towncrier (yet), so we have to just write to the changelog for now.


filipi87 (approved):
LGTM!

PR #252, opened 2026-04-15
Remove references to OpenAILLMContext for pipecat 1.0 compatibility

DominicStewart:
pipecat 1.0 removed OpenAILLMContext (along with OpenAILLMContextFrame and OpenAILLMContext.from_messages). Both manager.py and adapters.py imported it purely for use in `OpenAILLMContext | LLMContext` type-hint unions, which made the modules unimportable on pipecat 1.0 with:

  ModuleNotFoundError: No module named
  'pipecat.processors.aggregators.openai_llm_context'

This drops the now-dead import and trims the type-hint unions to `LLMContext`. No behavior change; the `generate_summary` isinstance fallback is left in place (it's unreachable on 1.0 but harmless, and keeping it avoids churning existing tests).

All 88 tests pass against pipecat-ai 1.0.0.


markbackman:
Thanks for submitting 🙏 

I'm working on a PR to remove deprecations, which will include this, so I'll close it to avoid merge conflicts.

PR #256, opened 2026-04-15
Clean up types

markbackman:


src/pipecat_flows/actions.py
```diff
@@ -138,7 +138,7 @@ def _register_action(self, action_type: str, handler: Callable) -> None:
         self._action_handlers[action_type] = handler
         logger.debug(f"Registered handler for action type: {action_type}")
 
-    async def execute_actions(self, actions: Optional[List[ActionConfig]]) -> None:
+    async def execute_actions(self, actions: list[ActionConfig] | None) -> None:
```
markbackman, line 141: This is the new convention for optionals in newer codebases. Thoughts on whether we want to go this route in general? @aconchillo

src/pipecat_flows/__init__.py
```diff
@@ -39,6 +40,7 @@
     # Flow Manager
     "FlowManager",
     # Types
+    "ActionConfig",
```
markbackman, line 43: This type was missing and should have been public.

aconchillo (approved):

src/pipecat_flows/actions.py
aconchillo: Yes, I think we should start doing this. Using `typing` package was a thing from 3.9 and union operator introduced in 3.10. So we should probably modernize everything.

PR #263, opened 2026-04-17
Add pyright, fix issues

markbackman:
## Summary

Configures pyright for the library, fixes the real type errors it surfaces, and uses the resulting visibility to align the flows type surface with Pipecat upstream. No breaking runtime or static-type changes; this is a MINOR-safe release.

The central through-line: our public types were tighter than Pipecat's runtime contract without enforcing anything. Results coming out are `Any` end-to-end in Pipecat; our `FlowResult` narrowing was purely advisory and caused legitimate user handlers (e.g. the quickstart) to fail type-checking. This PR loosens that to match upstream, deprecates the patterns that forced strict shapes, and lays out a migration path to 2.0.0.

## Setup

- `pyrightconfig.json` at repo root, `basic` mode, scoped to `src/`.
- Examples, tests, and build dirs excluded (keeps the IDE quiet too).
- Pyright added to `scripts/fix-ruff-and-typecheck.sh`.

## Deprecations (removed in 2.0.0)

- `ZeroArgFunctionHandler` and `LegacyFunctionHandler` (0-arg and 1-arg function handler shapes). Canonical signature is `(args: FlowArgs, flow_manager: FlowManager)`. Runtime `DeprecationWarning` fires once per `FlowManager` instance per shape.
- `LegacyActionHandler` (single-argument action handler shape). Canonical signature is `(action: dict, flow_manager: FlowManager)`. Runtime warning fires once per `ActionManager` instance.
- `FlowResult`. No longer referenced by any handler type in the library. Define your own `TypedDict` or return any JSON-serializable value.

## Changed

- Function handler return types widened to `Awaitable[Any]` and `ConsolidatedFunctionResult`'s first slot widened to `Any`, matching Pipecat's `FunctionCallResultCallback` contract. Strictly additive: handlers returning `FlowResult` still validate, and handlers returning raw strings or other JSON-serializable values (e.g. the quickstart) are now type-clean.

## Fixed

- `FlowsDirectFunction` redefined as `Callable[..., Awaitable[ConsolidatedFunctionResult]]` instead of a `Protocol` with `**kwargs: Any`. Python's Protocol system can't structurally match concrete named parameters against `**kwargs`, so every direct function users wrote (`async def f(flow_manager, llm: str)`) surfaced a pyright error. Runtime validation is unchanged, handled by `FlowsDirectFunctionWrapper.validate_function`.

## Deferred to 2.0.0

- `FlowArgs` will widen from `dict[str, Any]` to `Mapping[str, Any]` to align with Pipecat's typing of `FunctionCallParams.arguments`. For now we convert via `dict(params.arguments)` at the one internal call site, which also gives handlers a defensive copy. Users who mutate args should pin their handler annotation to `dict[str, Any]` explicitly; users who only read will be unaffected.

## Examples

- Dropped `FlowResult` subclassing across all examples; subclasses are now plain `TypedDict` with `status: str` declared explicitly where used.
- Aligned every example to use `NodeConfig(...)` constructor form (some had used dict literals).
- `warm_transfer.py`: narrowed `flow_manager.transport` to `DailyTransport` via `assert isinstance(...)` in the three handlers that need it, widened `get_*_participant_id` return types to `str | None`, and added an `agent_participant_id` truthy check to avoid passing `None` as a dict key.

## Test plan

- [x] `uv run pyright` clean (0 errors).
- [x] `uv run ruff check .` clean.
- [x] `uv run pytest tests/` passes (71 tests).
- [x] `uv run towncrier build --draft --version Unreleased` renders cleanly with five fragments across Changed / Deprecated / Fixed sections.
- [x] Quickstart example (`examples/quickstart/hello_world.py`) type-checks without errors even though its handler returns `tuple[str, NodeConfig]`.


src/pipecat_flows/types.py
```diff
@@ -61,6 +59,16 @@ class FlowResult(TypedDict, total=False):
 FlowArgs = dict[str, Any]
```
markbackman, line 59: I'd like to make this:
```
FlowArgs = Mapping[str, Any]
```

this would match Pipecat. But, it's breaking. We can wait to introduce this in 2.0.0. Or, do you have other ideas about how to transition?

src/pipecat_flows/types.py
```diff
@@ -39,19 +38,18 @@
 
 
 class FlowResult(TypedDict, total=False):
-    """Base type for function results.
+    """Optional convention TypedDict for ``status``/``error`` results.
```
markbackman, line 41: This just isn't worth it. Pipecat supports `Any` for the `result_callback`, so we should just follow suit. Developers can just use a `TypedDict` if they want.

src/pipecat_flows/types.py
```diff
@@ -70,69 +78,97 @@ class FlowResult(TypedDict, total=False):
...
 - doing some work
 - specifying the next node to transition to after the work is done
+
+The first tuple element is the function-call result delivered to the LLM.
+Any JSON-serializable value is accepted (matching Pipecat's upstream
+``FunctionCallResultCallback`` contract). Pass ``None`` to signal a
+transition-only handler; FlowManager substitutes an acknowledgement result.
+"""
+
+ZeroArgFunctionHandler = Callable[[], Awaitable[Any]]
```
markbackman, line 94: Adding this type to satisfy the type checker, but also planning to remove it and `LegacyFunctionHandler` in 2.0.0, aligning on the `FlowFunctionHandler`.

I would have done this with 1.0.0 but we didn't deprecate it 🤷.

src/pipecat_flows/types.py
```diff
@@ -70,69 +78,97 @@ class FlowResult(TypedDict, total=False):
...
 
-FunctionHandler = LegacyFunctionHandler | FlowFunctionHandler
-"""Union type for function handlers supporting both legacy and modern patterns."""
+FunctionHandler = ZeroArgFunctionHandler | LegacyFunctionHandler | FlowFunctionHandler
+"""Union type for function handlers supporting 0-arg, legacy, and modern patterns."""
 
 
-class FlowsDirectFunction(Protocol):
-    """Protocol for "direct" functions with automatic metadata extraction.
+FlowsDirectFunction = Callable[..., Awaitable[ConsolidatedFunctionResult]]
```
markbackman, line 138: See the commit about this one. Claude proposed this solution to resolve a type issue.

changelog/263.deprecated.md
```diff
@@ -0,0 +1 @@
+- 0-argument and 1-argument (legacy) function handlers are deprecated and will be removed in 2.0.0. Update handlers to accept `(args: FlowArgs, flow_manager: FlowManager)`. A `DeprecationWarning` is emitted once per `FlowManager` instance the first time each deprecated shape is invoked. The `ZeroArgFunctionHandler` type is now exported so the currently supported handler shapes are fully described by the public type surface.
```
kompfner, line 1: I'm liking all the deprecations you're introducing. They're overdue—Flows was getting a little bit "littered" with past patterns.

src/pipecat_flows/types.py
kompfner: Works for me! I'm not sure users were really benefiting from editor type help (e.g. autocomplete) much the way things were before anyway.

kompfner (approved):

src/pipecat_flows/types.py
markbackman: Note to ourselves: we'll want to follow the same pattern in Pipecat for the direct function types.

PR #270, opened 2026-05-12
Re-register function handlers when a node redeclares them

### LINKED ISSUE #269: Re-declaring a function in a new node silently keeps the previous node's handler bound
Hi 👋

When two adjacent nodes declare a function with the same name but different
handlers, the second node's handler is silently ignored. The tool schema sent
to the LLM is updated correctly (description, parameters), but the closure
that actually fires belongs to the first node. No log, no warning, no exception.

It only triggers on adjacent A→B nodes — any intermediate node that doesn't
re-declare the name resets `_current_functions` and B will register fresh.

## Where it lives

`src/pipecat_flows/manager.py:544` — the dedup guard in `_register_function`:

```python
if name not in self._current_functions:
    transition_func = await self._create_transition_func(name, handler)
    self._llm.register_function(name, transition_func, ...)
    new_functions.add(name)
```

In _set_node (manager.py:638-647), tools.append(schema) runs
unconditionally, so node B's schema does reach the LLM — only the handler
binding stays stale.

Repro

Two nodes declare go with different handlers. After setting node B,
register_function has been called once, not twice:

```python
await flow_manager.set_node_from_config({  # node A
    "task_messages": [...],
    "functions": [FlowsFunctionSchema(name="go", ..., handler=handler_a)],
})
await flow_manager.set_node_from_config({  # node B
    "task_messages": [...],
    "functions": [FlowsFunctionSchema(name="go", ..., handler=handler_b)],
})
```
# Expected: register_function called twice with name="go".
# Observed on main (d00c7f6): called once. handler_b never bound.

Suggested fix

Drop the guard and always call self._llm.register_function(...). Pipecat-core
already handles re-registration cleanly:

- LLMService.register_function is a pure dict upsert by name.
- _run_function_call re-resolves the registry at execute time, per its own
comment: "Re-resolve the registry item at execution time. The function may
have been unregistered between queuing and execution."

Verified locally: repro passes, no regressions in tests/.

Related

- #230 — adjacent (stale handlers across nodes). Redirected to pipecat core.
- #232 — drafted a structural fix (_current_functions as Dict[...],
always-replace). Closed not merged.
- pipecat#4301 / #4404 — what landed in core. Different slice of the problem.

PR with a one-line fix + regression test coming shortly.

Thanks!

madaram:
## Summary

- Drop the `name not in self._current_functions` guard in
  `FlowManager._register_function` so that handlers are re-registered with the
  LLM each time a node redeclares a function.
- Add `test_redeclared_function_rebinds_new_handler` in `tests/test_manager.py`
  asserting the new handler wins after re-declaration.

Fixes #269.

## Background

When two adjacent nodes declared the same function name with different
handlers, the OpenAI tool schema reaching the LLM was correctly updated to the
second node's version, but the handler closure stayed bound to the first
node's. Result: the LLM called the function expecting node B's behaviour, but
node A's transition fired — silently, no warning.

## Test plan

- [x] New regression test fails on `main` and passes after the fix.
- [x] `uv run pytest tests/` — 61 tests pass locally
      (`test_context_strategies.py` skipped — local env lacks the `google-genai`
      extra; behaviourally unrelated to this change).
- [x] `uv run ruff check .` — clean
- [x] `uv run ruff format --check .` — clean
- [ ] Changelog fragment to be added in a follow-up commit once the PR number
      is assigned.

### DIFF
--- a/changelog/270.fixed.md
+++ b/changelog/270.fixed.md
@@ -1,0 +1,1 @@
+- Fixed an issue where redeclaring a function in a new node silently kept the previous node's handler bound (#269).
--- a/src/pipecat_flows/manager.py
+++ b/src/pipecat_flows/manager.py
@@ -527,7 +527,11 @@ class FlowManager:
         cancel_on_interruption: bool = True,
         timeout_secs: float | None = None,
     ) -> None:
-        """Register a function with the LLM if not already registered.
+        """Register a function with the LLM.
+
+        Always re-registers so a node redeclaring a function with a different
+        handler replaces the previous binding. ``LLMService.register_function``
+        is a dict upsert by name, so re-registration is idempotent.
 
         Args:
             name: Name of the function to register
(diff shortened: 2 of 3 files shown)


markbackman (approved):
LGTM! Thanks for finding and fixing this one 🙏

PR #272, opened 2026-05-27
Resolve `ConsolidatedFunctionResult` without requiring `NodeConfig` import

timofey-TK:
## Summary

Fixes #271.

`ConsolidatedFunctionResult` is a public type alias users are expected to use as the return annotation of "direct" functions (per `FlowsDirectFunction` / `FlowsDirectFunctionWrapper`). In `1.1.0` it was defined as:

```python
# src/pipecat_flows/types.py
ConsolidatedFunctionResult = tuple[Any, "NodeConfig | None"]
```

The string forward reference exists because `NodeConfig` is defined later in the same file. When `BaseDirectFunctionWrapper` introspects a user-defined direct function, it calls `get_type_hints(func)`, which resolves string forward references against the **user's** module globals — not `pipecat_flows.types`. Unless the user separately imports `NodeConfig`, this raises:

```
NameError: name 'NodeConfig' is not defined
```

In practice the failure surfaces deep inside `FlowManager._set_node` and stalls the flow: the function call itself succeeds and the transition fires, but the next node never activates.

## Fix

Reorder `types.py` so `NodeConfig` is defined *before* `ConsolidatedFunctionResult`. The alias can then drop the string quotes:

```python
ConsolidatedFunctionResult = tuple[Any, NodeConfig | None]
```

Now the alias is fully resolved at module import time, so `get_type_hints()` on user code doesn't need to look up `NodeConfig` in the caller's globals.

`NodeConfig` retains string forward references for `FlowsFunctionSchema | FlowsDirectFunction` (which are still declared later in the file), but those references are only resolved by `get_type_hints(NodeConfig)`, which uses this module's globals — and the names do exist there.

A few other small definitions (`LegacyActionHandler`, `FlowActionHandler`, `ActionConfig`, `ContextStrategy`, `ContextStrategyConfig`) were moved up alongside `NodeConfig` because `NodeConfig` depends on them. The rest of the file is unchanged.

## Test plan

- [x] Added two regression tests in `tests/test_flows_direct_functions.py`:
  - `test_get_type_hints_resolves_without_nodeconfig_import` — runs the exact repro from the issue.
  - `test_direct_function_wrapper_accepts_consolidated_return_type` — verifies `FlowsDirectFunctionWrapper.validate_function` / construction succeed for a direct function annotated with `-> ConsolidatedFunctionResult`, since the wrapper internally calls `get_type_hints`.
- [x] `uv run pytest tests/` — all 74 tests pass.
- [x] `uv run ruff format --check .` and `uv run ruff check .` — clean.
- [x] Confirmed the minimal repro from the issue now succeeds without importing `NodeConfig`.

### DIFF
--- a/changelog/272.fixed.md
+++ b/changelog/272.fixed.md
@@ -1,0 +1,1 @@
+- Fixed an unresolved `NodeConfig` forward reference in `ConsolidatedFunctionResult` that caused `get_type_hints()` (and therefore Pipecat's `BaseDirectFunctionWrapper`) to raise `NameError: name 'NodeConfig' is not defined` on direct functions annotated with `-> ConsolidatedFunctionResult`, stalling flows at node-setup time (#271).
--- a/src/pipecat_flows/types.py
+++ b/src/pipecat_flows/types.py
@@ -78,89 +78,6 @@ Example::
     }
 """
 
-ConsolidatedFunctionResult = tuple[Any, "NodeConfig | None"]
-"""Return type for "consolidated" functions.
-
-Return type for "consolidated" functions that do either or both of:
-- doing some work
-- specifying the next node to transition to after the work is done
-
-The first tuple element is the function-call result delivered to the LLM.
-Any JSON-serializable value is accepted (matching Pipecat's upstream
-``FunctionCallResultCallback`` contract). Pass ``None`` to signal a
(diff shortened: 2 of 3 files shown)


markbackman (approved):
LGTM! Thanks for catching this 🙌

PR #275, opened 2026-05-29
Add classifier_routing example: route nodes without an LLM call

jamsea:
## Summary

Adds a new example, `examples/classifier_routing.py`, showing how to route between Flow nodes using a classifier instead of an LLM call.

Some node decisions are trivial: a verification step that only needs a yes/no, a menu where the user picks one of a few known options. Paying for an LLM inference to make that decision adds latency and cost you don't need. This example demonstrates the idiomatic Flows pattern for making those decisions with a plain classifier (a regex here, but the `classify()` function is the seam to swap in a rules engine or a small binary classifier), while reserving the LLM only for inputs the classifier can't confidently handle.

## How it works

The example is a tiny age-verification bot:

- The `verify` node uses `respond_immediately=False`, so entering it triggers no LLM completion. It speaks its canned question via a `tts_say` pre-action and then waits for the user.
- In a normal pipeline, the user's turn-end emits an `LLMContextFrame` that the LLM service consumes and responds to. That is the LLM call we want to skip.
- `ClassifierRouter` is a custom `FrameProcessor` placed between the user context aggregator and the LLM. While in the `verify` node it classifies the user's answer and either:
  - routes deterministically (`yes` -> confirmed, `no` -> declined) via `set_node_from_config(...)` and drops the `LLMContextFrame` so the LLM never runs, or
  - passes the frame through to the LLM when it can't classify the input, letting the LLM improvise an answer (the follow-up / off-topic case).
- The terminal `confirmed`/`declined` nodes also avoid the LLM: they speak a canned line and end the call with an `end_conversation` pre-action.

`ClassifierRouter` drops exactly one frame type (`LLMContextFrame`), and only in the deterministic branch, because that frame is precisely the LLM-run trigger. Every other frame is always pushed through.

## Changes

- New: `examples/classifier_routing.py`
- `examples/README.md`: listed the new example under Implementation Patterns
- Changelog fragment

## Testing

- `ruff format` + `ruff check`: clean
- Offline smoke test: module imports, `classify()` returns the expected yes/no/None across cases, all three node builders produce valid `NodeConfig`s, and the router instantiates
- Run it: `uv run examples/classifier_routing.py` (needs `CARTESIA_API_KEY` + `GOOGLE_API_KEY`), then open http://localhost:7860/client. Saying "yes"/"no" routes with no LLM call between turn-end and the reply; asking a question instead falls through to the LLM.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,5 +1,10 @@
 {
   "attribution": {
     "commit": ""
+  },
+  "permissions": {
+    "allow": [
+      "Bash(uv run *)"
+    ]
   }
 }
--- a/changelog/275.added.md
+++ b/changelog/275.added.md
@@ -1,0 +1,1 @@
+- Added `classifier_routing.py` example showing how to route between Flow nodes using a classifier/regex instead of an LLM call, with canned responses for trivial decisions and an LLM fallback for unmatched input.
--- a/examples/README.md
+++ b/examples/README.md
@@ -22,6 +22,7 @@ See [quickstart/README.md](./quickstart/README.md) for a beginner-friendly intro
 - `food_ordering_direct_functions.py` - Food ordering using direct function registration
 - `restaurant_reservation_direct_functions.py` - Reservation system using direct function registration
 - `llm_switching.py` - Switching between different LLM providers during conversation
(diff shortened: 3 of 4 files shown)


jamsea:
Here's a demo:
https://www.loom.com/share/6fcfcfd0aed448948f03ef3b37ebe3b8

PR #277, opened 2026-05-29
Migrate to Pipecat 1.3.0 worker model

markbackman:
## Summary

- Migrates Pipecat Flows to the [Pipecat 1.3.0 worker model](https://github.com/pipecat-ai/pipecat/pull/4493), which reframes the central abstraction from a single *pipeline task* to a *worker* a runner can host alongside others (`PipelineTask` → `PipelineWorker`, `PipelineRunner` → `WorkerRunner`).
- `FlowManager` now accepts a `worker` (`PipelineWorker`) argument and exposes a `worker` property.
- Examples are fully modernized to `PipelineWorker` + `WorkerRunner` using `runner.add_workers(worker)` + `runner.run()`.

## Backwards compatibility

Existing code keeps working — this is additive:

- `FlowManager(task=...)` still works but emits a `DeprecationWarning`; passing both `worker` and `task`, or neither, raises `ValueError`.
- The `FlowManager.task` property still resolves to the worker (deprecated, warns once).
- Pipecat itself keeps `PipelineTask`/`PipelineRunner` as deprecated aliases, so a `PipelineTask` is a `PipelineWorker`.

## Notes for reviewers

- Bumps the dependency to `pipecat-ai>=1.3.0,<2`.
- Internal `_task` → `_worker` in `FlowManager` and `ActionManager`.
- Tests were updated for the now-required `FunctionCallParams.pipeline_worker` field (a separate 1.3.0 change), and new coverage was added for the `worker`/`task` constructor and property behavior.

## Testing

- `uv run pytest tests/` — 75 passed
- `uv run ruff check .` / `uv run ruff format --check .` — clean
- Importing a migrated example under `python -W error::DeprecationWarning` produces no warnings, confirming examples use only the new API


src/pipecat_flows/manager.py
```diff
@@ -88,19 +88,25 @@ class FlowManager:
     def __init__(
         self,
         *,
-        task: PipelineTask,
         llm: LLMService | LLMSwitcher,
         context_aggregator: Any,
+        worker: PipelineWorker | None = None,
+        task: PipelineWorker | None = None,
```
markbackman, line 94: `PipelineTask` is a subclass of `PipelineWorker`. Makes this cleaner in that we don't have to use a deprecated path.

src/pipecat_flows/actions.py
```diff
@@ -85,15 +85,15 @@ class ActionManager:
     Custom actions can be registered using register_action().
     """
 
-    def __init__(self, task: PipelineTask, flow_manager: "FlowManager"):
+    def __init__(self, worker: PipelineWorker, flow_manager: "FlowManager"):
```
markbackman, line 88: ActionManager is not a public class in __init__, so we can just migrate to `worker`.

aconchillo (approved):

aconchillo:
LGTM!

PR #280, opened 2026-06-04
Add multi-worker handoff example combining Flows with the worker framework

aconchillo:
## Summary

Adds a new example, `examples/multi_worker_handoff.py`, demonstrating how Pipecat Flows composes with Pipecat's multi-worker framework (the worker model introduced with Pipecat 1.3.0). It ports the idea from the now-archived `pipecat-subagents` `FlowsAgent` example onto the native worker API.

Three workers share a single bus:

- **Main worker** — owns the transport (STT/TTS) and the single shared `LLMContext`; bridges user/assistant frames onto the bus via `BusBridgeProcessor`. It runs no LLM itself.
- **Router worker** — a plain `LLMWorker` for open-ended chit-chat, with `@tool` `transfer_to_reservation` / `end_conversation`.
- **Reservation worker** — a structured Flows conversation (party size → time → availability → confirm), with a `transfer_to_router` escape hatch at every node.

Only one worker is active at a time; hand-offs (via `activate_worker(..., deactivate_self=True)`) are seamless. All workers speak into the main worker's shared `context_aggregator`.

The reservation worker is built by `build_reservation_worker(...)` returning a plain `PipelineWorker` (same style as the `sensor-controller` example) — no subclass. A `FlowManager` is wired onto the worker and the flow is initialized/resumed from the worker's `on_activated` event handler.


markbackman (approved):
LGTM! Thanks for making this update 🙌

PR #281, opened 2026-06-04
Add append_text_to_context for TTS actions; preserve initial-node context

kompfner:
## Summary

A few related changes for how Pipecat Flows writes programmatically-generated TTS output to the LLM context.

1. The `tts_say` and `end_conversation` actions now append their spoken text to the LLM context by default (`append_text_to_context` defaults to `True`)
2. The initial flow node now follows its context strategy (default `APPEND`) instead of always replacing the context. With the default strategy, context written before the node's first inference — for example by a `tts_say` pre-action — is now preserved.
3. Added an `append_text_to_context` option to the `tts_say` and `end_conversation` actions, controlling whether the spoken text is written to the LLM context.

## Testing

`uv run pytest tests/` — all green against the released `pipecat-ai` (1.3.0) as well as the unreleased tip of the Pipecat `main` branch. Also validated live against a local example exercising `tts_say` as pre- and post-actions and `end_conversation`, toggling `append_text_to_context` across `True`/`False`/unset and printing the resulting context.

### DIFF
--- a/changelog/281.added.md
+++ b/changelog/281.added.md
@@ -1,0 +1,1 @@
+- Added an `append_text_to_context` option to the `tts_say` and `end_conversation` actions, controlling whether the spoken text is written to the LLM context.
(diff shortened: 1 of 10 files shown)


filipi87 (approved):
LGTM!

markbackman (approved):
LGTM!

PR #282, opened 2026-06-10
Rename @flows_direct_function to @flows_tool_options

markbackman:
## Summary

- Added `@flows_tool_options(cancel_on_interruption=..., timeout_secs=...)` as the primary decorator for configuring a Flows direct function's call options. It is the Flows-flavored counterpart to Pipecat's `@tool_options` and is built on it, so the metadata now lives under the `_pipecat_*` attributes the wrapper reads.
- Deprecated `@flows_direct_function` — it is now an alias of `@flows_tool_options` that emits a `DeprecationWarning` and delegates to it. The alias will be removed in a future version.
- Bumped the minimum `pipecat-ai` to `1.4.0` (core dependency and the examples extra), where `@tool_options` and the `timeout_secs` naming land.

The point of the rename is naming alignment: matching Pipecat's `@tool_options` makes it clearer that the decorator configures a direct function's *call options*, rather than implying it's what makes a function a "direct function."

## ⚠️ Blocked on Pipecat 1.4.0

`@tool_options` and the `timeout_secs` decorator argument only exist on Pipecat's `main` (unreleased). **This PR cannot pass CI or be merged until `pipecat-ai` 1.4.0 is published to PyPI** — CI installs the released `pipecat-ai`, which still lacks `tool_options`. Verified locally against the Pipecat checkout via `PYTHONPATH`.

## Breaking Changes

- None at runtime. `@flows_direct_function` keeps working (with a deprecation warning). The minimum `pipecat-ai` requirement rises to 1.4.0.

## Testing

```
uv run pytest tests/test_flows_direct_functions.py tests/test_manager.py
```

(Run against a Pipecat 1.4.0+ environment.) Includes a new test that `@flows_direct_function` warns but still attaches the options, and that `@flows_tool_options` does not warn.


changelog/282.added.md
```diff
@@ -0,0 +1 @@
+- Added the `@flows_tool_options(cancel_on_interruption=..., timeout_secs=...)` decorator for configuring a Flows direct function's call options. It is the Flows-flavored counterpart to Pipecat's `@tool_options` (and is built on it), replacing `@flows_direct_function` with a name that makes clearer that it configures call options. Requires `pipecat-ai>=1.4.0`.
```
kompfner: Nit: does it matter to the reader here that it requires a newer version of Pipecat? They'll know, upon bumping their version of Pipecat Flows, that the newer Pipecat is required.

src/pipecat_flows/types.py
```diff
@@ -369,13 +370,19 @@ def to_function_schema(self) -> FunctionSchema:
...
 
 
-def flows_direct_function(
+def flows_tool_options(
     *, cancel_on_interruption: bool = False, timeout_secs: float | None = None
 ) -> Callable[[Callable], Callable]:
-    """Decorator to attach additional metadata to a Pipecat direct function.
+    """Configure a Flows direct function's call options.
+
+    This decorator is optional; use it to override the defaults for a Flows
```
kompfner, line 378: Nit: everything after "use it to override the defaults for a Flows direct function" feels like it might be TMI to the reader.

tests/test_flows_direct_functions.py
```diff
@@ -275,7 +276,7 @@ async def my_function(flow_manager: FlowManager):
     def test_cancel_on_interruption_can_be_set_to_false(self):
         """Test that cancel_on_interruption can be set to False via decorator."""
 
-        @flows_direct_function(cancel_on_interruption=False)
+        @flows_tool_options(cancel_on_interruption=False)
```
kompfner, line 279: Reads much better!

kompfner (approved):

src/pipecat_flows/types.py
markbackman: Fixed.

changelog/282.added.md
markbackman: Fixed.

PR #283, opened 2026-06-17
Adopt Pipecat's function-calling best practices

kompfner:
## What

Updates Pipecat Flows to use Pipecat core's function-handler auto-registration (new in `pipecat-ai` 1.4.0) instead of calling `LLMService.register_function` directly.

## Changes

- **Auto-registration.** Flows no longer calls `LLMService.register_function`, relying on Pipecat's new auto-registration mechanism instead.
- **`FlowsFunctionSchema.handler` is now required.** Its optionality was a holdover from the removed `transition_to`/`transition_callback` fields; a handler-less schema has had no purpose since.
- Inlined the now-trivial `LLMAdapter.format_functions` into `FlowManager`. We're now one step closer to being able to remove `LLMAdapter` altogether. It's somewhat of a relic from when we had to do a bunch of LLM-specific adapting.
- **Examples now lead with direct functions.** Direct functions are the standard across all examples; one `FlowsFunctionSchema` example (`food_ordering_advanced_functionschema.py`) is kept to show the explicit-schema approach (e.g. `enum` / min-max property constraints).

## Still to do (this PR)

- [x] Update examples to lead with direct functions.


changelog/283.changed.md
```diff
@@ -0,0 +1 @@
+- ⚠️ `FlowsFunctionSchema.handler` is no longer typed as optional. In practice, it wasn't optional anyway: the optionality was a holdover from the removed `transition_to`/`transition_callback` fields, which once let a function transition without a handler. Transitions now happen via a handler that returns `(result, next_node)`. Constructing a `FlowsFunctionSchema` without a `handler` now raises an error.
```
kompfner, line 1: Reviewers: please note this. Do let me know if you think I'm mistaken about the claim that, in practice, `handler` was already mandatory. I couldn't find any reason to keep it optional, and making it mandatory is what allows us to cleanly do the main work of this PR (letting Pipecat auto-register all functions).

markbackman: Nice catch. Yes, this makes sense. And, you're right about the historical piece. I think this should have been marked as mandatory in migrating to 1.0.0.

src/pipecat_flows/types.py
```diff
@@ -333,8 +333,8 @@ async def f(flow_manager: FlowManager, **params) -> ConsolidatedFunctionResult:
 class FlowsFunctionSchema:
```
markbackman, line 333: What's interesting is that a `FlowsFunctionSchema` is now essentially a `FunctionSchema` decorated with `@tool_options`.

Because of the handler type difference, these need to stay distinct, but it's great to see how they've converged. Very little gap between Pipecat and Pipecat Flows now 🎉

markbackman (approved):
Very nice clean up. It's great to see the complexity fall away.

On the handler change, it's technically breaking, but in practice harmless. A handler-less FlowsFunctionSchema was already a no-op (just logged a warning and acknowledged the call) since 1.0.0. Better to error than it fail silently.

I'm happy to proceed 👍

PR #285, opened 2026-06-18
Use a direct function in the quickstart example

kompfner:
## What

Defines the quickstart's `record_favorite_color` as a direct function instead of a `FlowsFunctionSchema`, bringing the Hello World example in line with the direct-functions-first standard the other examples adopted in #283.

This matches the updated quickstart guide ([pipecat-ai/docs#928](https://github.com/pipecat-ai/docs/pull/928)).


markbackman (approved):
Ship it!

